PR #14, opened 2024-02-13
SM-579 Get User by Eth Addr User Service

lauradeng23:
Add handler to get a user by eth address rather than user id

update protos/methods/controllers


internal/api/handlers.go
```diff
@@ -26,6 +26,13 @@ func GetUserID(c *fiber.Ctx) string {
 	return userID
 }
 
+func GetUserEthAddr(c *fiber.Ctx) string {
```
JamesReate, line 29: let's have this return a crypto Address, eg. return `common.HexToAddress(ethAddr)`

internal/controllers/token_exchange.go
```diff
@@ -85,28 +85,23 @@ func (t *TokenExchangeController) GetDeviceCommandPermissionWithScope(c *fiber.C
 		return fiber.NewError(fiber.StatusInternalServerError, "Could not connect to blockchain node")
 	}
 
-	userID := api.GetUserID(c)
-	user, err := t.usersService.GetUserByID(c.Context(), userID)
+	e := api.GetUserEthAddr(c)
```
JamesReate, line 88: let's still get the userID - latest I heard is not all JWT's will have an ethereum_address claim. So if there's no ethereum address do the lookup by userId

internal/services/users_service.go
```diff
@@ -46,3 +47,15 @@ func (u *usersService) GetUserByID(ctx context.Context, userID string) (*pb.User
 		Id: userID,
 	})
 }
+
+func (u *usersService) GetUserByEthAddr(ctx context.Context, ethAddr string) (*pb.User, error) {
```
JamesReate, line 51: have this take in an eth addr object from the go-ethereum library. Below casting string to bytes won't render the expected output `[]byte(ethAddr)` - this will give you the byte representation of the string not the equivalent hex to bytes.

lauradeng23: done!

internal/controllers/token_exchange.go
lauradeng23: done

internal/api/handlers.go
lauradeng23: doneee

JamesReate (approved):
I think this should handle the scenarios we had in mind and not cause breaking changes since we still check for the userId

internal/controllers/token_exchange.go
```diff
@@ -85,28 +85,34 @@ func (t *TokenExchangeController) GetDeviceCommandPermissionWithScope(c *fiber.C
...
 		if s, ok := status.FromError(err); ok && s.Code() == codes.NotFound {
-			t.logger.Debug().Str("userId", userID).Msg("User not found.")
-			return fiber.NewError(fiber.StatusForbidden, "User not found!")
+			t.logger.Debug().Str("ethAddr", ethAddr.Hex()).Msg("Ethereum address not found.")
+			return fiber.NewError(fiber.StatusForbidden, "Ethereum address not found!")
 		}
-		t.logger.Error().Str("userID", userID).Msg("Users api unavailable!")
+		t.logger.Error().Str("ethAddr", ethAddr.Hex()).Msg("Users api unavailable!")
 		return fiber.NewError(fiber.StatusInternalServerError, err.Error())
 	}
```
elffjs, line 110: What is the point of this branch now?

internal/services/users_service.go
```diff
@@ -46,3 +48,15 @@ func (u *usersService) GetUserByID(ctx context.Context, userID string) (*pb.User
 		Id: userID,
 	})
 }
+
+func (u *usersService) GetUserByEthAddr(ctx context.Context, ethAddr common.Address) (*pb.User, error) {
+	client, conn, err := u.getUsersServiceGrpcConnection()
```
elffjs, line 53: It's unfortunate that we start a new connection for every request. I know you didn't start this pattern but it is unfortunate.

internal/controllers/token_exchange.go
```diff
@@ -85,28 +85,34 @@ func (t *TokenExchangeController) GetDeviceCommandPermissionWithScope(c *fiber.C
 		return fiber.NewError(fiber.StatusInternalServerError, "Could not connect to blockchain node")
 	}
 
-	userID := api.GetUserID(c)
-	user, err := t.usersService.GetUserByID(c.Context(), userID)
+	ethAddr := api.GetUserEthAddr(c)
```
elffjs, line 88: Could we add a test? I know the original work doesn't have any and that's awful, but we ought to be able to test that we fall back and all this.

I'm okay to be overridden on this.

JamesReate: yeah i think that would be good too, I'll help Laura with that

lauradeng23: test is up and passing

internal/controllers/token_exchange_test.go
```diff
@@ -69,3 +84,16 @@ func buildRequest(method, url, body string) *http.Request {
 
 	return req
 }
+
+// authInjectorTestHandler injects fake jwt with sub
+func authInjectorTestHandler(ethAddr common.Address) fiber.Handler {
+	return func(c *fiber.Ctx) error {
+		token := jwt.NewWithClaims(jwt.SigningMethodHS256, jwt.MapClaims{
+			"ethereum_address": ethAddr.Hex(),
+			"nbf":              time.Now().Unix(),
```
elffjs, line 93: Ahh this is tough.

JamesReate:
ok, i think this test is pretty robust. will leave updating to go 1.21 and deps for future

PR #23, opened 2024-08-19
Feature/as 2920 check sacd when issuing privilege tokens

LorranSutter:


internal/controllers/token_exchange.go
```diff
@@ -92,6 +92,12 @@ func (t *TokenExchangeController) GetDeviceCommandPermissionWithScope(c *fiber.C
 		t.logger.Fatal().Err(err).Str("Contracts", pr.NFTContractAddress).Msg("Unable to initialize nft contract")
 		return fiber.NewError(fiber.StatusInternalServerError, "Could not connect to blockchain node")
 	}
+	s, err := t.ctmr.GetSacd(t.settings.ContractAddressSacd, client)
+	if err != nil {
+		t.logger.Fatal().Err(err).Str("Contracts", pr.NFTContractAddress).Msg("Unable to initialize nft contract")
```
elffjs, line 97: This fatal is probably going to tear down the entire app, so no point in returning anything really. We should probably take out the fatal.

internal/controllers/token_exchange.go
```diff
@@ -118,6 +124,15 @@ func (t *TokenExchangeController) GetDeviceCommandPermissionWithScope(c *fiber.C
 			return fiber.NewError(fiber.StatusForbidden, fmt.Sprintf("Address lacks privilege %d.", p))
 		}
 	}
+
+	res, err := s.HasPermissions(nil, common.HexToAddress(pr.NFTContractAddress), big.NewInt(pr.TokenID), *ethAddr, intArrayToBitArray(pr.Privileges, 256))
```
elffjs, line 128: What feels bad about this is that I can't tell someone which permissions they don't have.

internal/controllers/token_exchange.go
```diff
@@ -139,3 +154,16 @@ func (t *TokenExchangeController) GetDeviceCommandPermissionWithScope(c *fiber.C
 		Token: tk,
 	})
 }
+
+func intArrayToBitArray(indices []int64, length int) *big.Int {
+	bitArray := big.NewInt(0)
+
+	for _, index := range indices {
+		if index >= 1 && index <= int64(length) {
```
elffjs, line 162: Is this good enough? I think I need `2*index` to be less than `length`.

internal/contracts/sacd/Sacd.go
```diff
@@ -0,0 +1,530 @@
+// Code generated - DO NOT EDIT.
+// This file is a generated binding and any manual changes will be lost.
+
+package contracts
```
elffjs, line 4: Package should be `sacd` if the folder's named that.

LorranSutter: Multiprivilege is capital case. Do we have a style guide for that?

internal/controllers/token_exchange.go
LorranSutter: The param I passed above was 256. But I removed that, so doesn't matter anymore

internal/controllers/token_exchange.go
LorranSutter: Removed from multiprivilege too

There is also `Fatal` in the client initialization above. Does it make sense to remove there too? Not sure what would be the best way to handle that in this case.

internal/controllers/token_exchange.go
LorranSutter: I changed to query one by one

LorranSutter: I think none of that will be necessary once I include the sacd call in the NFT tbh

internal/contracts/sacd/Sacd.go
elffjs: This is just writing any Go code for the last 14 years. There are lots of good style guides but I'm not sure they ever call this out.

internal/controllers/token_exchange.go
elffjs: That was a mistake too.

PR #25, opened 2025-01-06
EX-3238-require-dev-license

Allyson-English:
Evaluate JWT in request to determine whether DIMO mobile is sending the request. If so, proceed as normal (no changes yet, but we will plan to add additional security here soon).

If the request is not coming from DIMO mobile, ensure that the request is coming from a valid dev license. We will do this by:
- checking that the subject of the JWT is a valid dev license


internal/controllers/identity_api.go
```diff
@@ -0,0 +1,106 @@
...
+	return &IdentityApiController{
+		logger:      logger,
+		identityUrl: settings.IdentityAPIURL,
+	}
+}
+
+func (i *IdentityApiController) isDevLicense(ctx context.Context, ethAddr common.Address) (bool, error) {
+	requestBody := map[string]any{
+		"query": `{
+			query ($clientId: clientId!) {
```
elffjs, line 32: I think this needs a name, maybe? I don't think `clientId!` is a type.

internal/controllers/identity_api.go
```diff
@@ -0,0 +1,106 @@
...
+		return false, fmt.Errorf("non-200 response from GraphQL API: %d, '%s'", resp.StatusCode, string(bodyBytes))
+	}
+
+	var respBody IdentityResponse
+	if err := json.Unmarshal(bodyBytes, &respBody); err != nil {
+		return false, fmt.Errorf("failed to unmarshal GraphQL response: %w", err)
+	}
+
+	if len(respBody.Errors) > 0 {
+		if respBody.Errors[0].Message == "sql: no rows in result set" {
```
elffjs, line 78: Oh dear let's not assume this shape.

elffjs: image

elffjs: Like that `database/sql` error coming through is an egg on our faces.

elffjs: Note also that—one might ask why we don't check identity for permissions. Why do we check the chain? Why keep doing this?

internal/controllers/token_exchange.go
```diff
@@ -116,6 +120,20 @@ func (t *TokenExchangeController) GetDeviceCommandPermissionWithScope(c *fiber.C
 		ethAddr = &e
 	}
 
+	aud := pr.Audience
```
elffjs, line 123: We want the audience from the _identity token_. Pretty much no one is passing in audience in the request body.

elffjs:
We are going to need a setting for the address of the mobile dev license

internal/api/handlers.go
```diff
@@ -38,6 +38,17 @@ func GetUserEthAddr(c *fiber.Ctx) *common.Address {
 	return &e
 }
 
+func GetClaimSubject(c *fiber.Ctx) *common.Address {
+	token := c.Locals("user").(*jwt.Token)
+	claims := token.Claims.(jwt.MapClaims)
+	subj, ok := claims["sub"].(string)
+	if !ok || !common.IsHexAddress(subj) {
```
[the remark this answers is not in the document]
Allyson-English: from what I've been understanding of the conversation, won't the ethereum_address represent something different? can we always assume that the ethereum address and the subject are interchangeable if the request is coming from a valid dev license?

internal/controllers/identity_api.go
Allyson-English: I updated this to just check for any amount of errors.. do you want me to update identity api to clean up that db error leaking through?

internal/controllers/token_exchange_test.go
```diff
@@ -70,14 +71,14 @@ func TestTokenExchangeController_GetDeviceCommandPermissionWithScope(t *testing.
 			permissionTokenRequest: &PermissionTokenRequest{
 				TokenID:            123,
 				Privileges:         []int64{4},
-				NFTContractAddress: "0x90c4d6113ec88dd4bdf12f26db2b3998fd13a144",
+				NFTContractAddress: "0x90C4D6113Ec88dd4BDf12f26DB2b3998fd13A144",
```
KevinJoiner, line 74: nothing wrong here but because I am lazy I often do this to get the correct check sum string in tests
``` go
common.HexToAddress("0x90c4d6113ec88dd4bdf12f26db2b3998fd13a144").String()
```

internal/config/settings.go
```diff
@@ -13,4 +13,5 @@ type Settings struct {
 	UsersAPIGRPCAddress      string `yaml:"USERS_API_GRPC_ADDRESS"`
 	ContractAddressWhitelist string `yaml:"CONTRACT_ADDRESS_WHITELIST"`
 	ContractAddressSacd      string `yaml:"CONTRACT_ADDRESS_SACD"`
+	IdentityURL              string `yaml:"IDENTITY_URL"`
```
KevinJoiner, line 16: Make sure you add this to `values.yaml` and `values-prod.yaml`

internal/middleware/valid_dev_license.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/DIMO-Network/token-exchange-api/internal/config"
+	"github.com/DIMO-Network/token-exchange-api/internal/services"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/gofiber/fiber/v2"
+	"github.com/golang-jwt/jwt/v5"
+	"github.com/rs/zerolog"
+)
+
+// MobileAppAudience Audience in DIMO mobile JWT
+const MobileAppAudience = "dimo-driver"
```
KevinJoiner, line 18: Does this need to be exported?

internal/middleware/valid_dev_license.go
```diff
@@ -0,0 +1,79 @@
...
+		), "*")
+
+		valid, err := idSvc.IsDevLicense(c.Context(), common.HexToAddress(subject))
+		if err != nil {
+			return err
+		}
+
+		if valid {
+			if subject != aud {
+				logger.Debug().Str("subject", subject).Str("audience", aud).Msg("developer jwt sub and aud do not match")
```
KevinJoiner, line 71: I don't think this will ever match from LIWD
`decoded(subject) == ethereum_address` and `audience == devLicense`

internal/middleware/valid_dev_license.go
```diff
@@ -0,0 +1,79 @@
...
+			return fiber.NewError(fiber.StatusBadRequest, err.Error())
+		}
+
+		subject := strings.Trim(string(
+			bytes.TrimSpace(
+				bytes.TrimSuffix(decoded, []byte("\x12\x04web3")),
+			),
+		), "*")
+
+		valid, err := idSvc.IsDevLicense(c.Context(), common.HexToAddress(subject))
```
KevinJoiner, line 64: Looks like the dev license is in the audience not the subject for the token I got yesterday
```json
{
  "iss": "https://auth.dimo.zone",
  "provider_id": "web3",
  "sub": "CioweDY4RDY4QjU1YjBmYzExRjRkMEQ5MTQzM0E3RjYyMTdmMkNDRDAyZTMSBHdlYjM",
  "aud": "0xf5ada890DA2E5582E38DF4648F9dAeE00e691199",
  "exp": 1737557012,
  "iat": 1736347412,
  "at_hash": "LZeyIdtsRuQR4Pw1pDtnRA",
  "email_verified": false,
  "ethereum_address": "0x68D68B55b0fc11F4d0D91433A7F6217f2CCD02e3"
}
```

internal/controllers/token_exchange_test.go
```diff
@@ -186,6 +187,75 @@ func TestTokenExchangeController_GetDeviceCommandPermissionWithScope(t *testing.
 	}
 }
 
+const (
+	developerAuthToken = `Bearer eyJhbGciOiJSUzI1NiIsImtpZCI6ImRkNTFkNDkwYjc1Y2VhOTNlMGI3YWI2YzcwODczNWVlN2FmZDBmMDgifQ.eyJpc3MiOiJodHRwczovL2F1dGguZGltby56b25lIiwicHJvdmlkZXJfaWQiOiJ3ZWIzIiwic3ViIjoiQ2lvd2VEWmxNMlk1Um1FME1VUTFOV1kzTWpZd05UQkRNelZsWVRaQlJtVmtOalZsTURZME1XSTVOVGNTQkhkbFlqTSIsImF1ZCI6IjB4NmUzZjlGYTQxRDU1ZjcyNjA1MEMzNWVhNkFGZWQ2NWUwNjQxYjk1NyIsImV4cCI6MTcyNDI0NDE3NCwiaWF0IjoxNzIzMDM0NTc0LCJhdF9oYXNoIjoiZVpzS2p5SzB0TGY2UkFNZkxKM1AydyIsImVtYWlsX3ZlcmlmaWVkIjpmYWxzZSwiZXRoZXJldW1fYWRkcmVzcyI6IjB4NmUzZjlGYTQxRDU1ZjcyNjA1MEMzNWVhNkFGZWQ2NWUwNjQxYjk1NyJ9.n7w63IvKTBqynVIggMCJAuty7P9nyCWugF0oxjipgzw9P7LvctzEXaheJmrWoP95QZJg9izaFWL2UoE4VpcnR4-_G6R2whZGV2aqlj8FQH1mQuznJZQyZUc6zKMi0wqedGEIYWBRI1zmXHy70_rXYnV4U4loPqKrXxXrhQ6oZWqCb9WxOdX5zf41LuYF6Ez2xk_jiciKxrvjoGtFsJK4fKhKRkzbO0i5IcdmQwrPEN75k8DxtYTHiYO8p_8BXY5Wej3lfEo6ZVtLumxfdkanILiOd-cY793Ru7sFvY6ObAsA9OLM-F1VmiRkCaHTaTK9t3DwPGmuDgduStFDLVX76Q`
```
KevinJoiner, line 191: Similar to other comment but this token is from `2024-08-21T08:42:54-04:00` and I think the format may have updated since

internal/services/identity_service.go
```diff
@@ -0,0 +1,124 @@
...
+	"fmt"
+	"io"
+	"net/http"
+
+	"github.com/DIMO-Network/token-exchange-api/internal/config"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/rs/zerolog"
+)
+
+//go:generate mockgen -source identity_service.go -destination mocks/identity_service_mock.go
```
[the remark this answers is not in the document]
Allyson-English: Its not getting used by any controller, it gets passed into the dev license validation middleware function

internal/middleware/valid_dev_license.go
Allyson-English: The subject can be base64 decoded into the developer license and it *should* match the ethereum address. I decode the string right above this call so that I'm passing in what we expect to be a valid developer license.

charts/token-exchange-api/values.yaml
```diff
@@ -31,6 +31,7 @@ env:
   USERS_API_GRPC_ADDRESS: users-api-dev:8086
   CONTRACT_ADDRESS_WHITELIST: 0x45fbcd3ef7361d156e8b16f5538ae36dedf61da8,0xA4ad0F9c722588910791A9BAC63ADbB365614Bc7
   CONTRACT_ADDRESS_SACD: '0x4E5F9320b1c7cB3DE5ebDD760aD67375B66cF8a3'
+    IDENTITY_URL: 'https://identity-api.dev.dimo.zone/query'
```
elffjs, line 34: Indent looks off

cmd/token-exchange-api/api.go
```diff
@@ -85,7 +89,7 @@ func startWebAPI(ctx context.Context, logger zerolog.Logger, settings *config.Se
 	// All api routes should be under v1
 	v1Route := app.Group("/v1")
 	// Token routes
-	tokenRoutes := v1Route.Group("/tokens", jwtAuth)
+	tokenRoutes := v1Route.Group("/tokens", jwtAuth, devLicenseMiddleware)
```
elffjs, line 92: I can never remember what the right order is.

I feel like this makes more sense, if it stays as middleware, on the handler?

Middleware for one endpoint has always seemed insane to me.

internal/controllers/token_exchange_test.go
elffjs: Does this matter? Are we doing string equality anywhere?

internal/middleware/valid_dev_license.go
```diff
@@ -0,0 +1,68 @@
...
+		if err != nil {
+			return err
+		}
+
+		if valid {
+			if !slices.Contains(aud, subject) {
+				logger.Debug().Str("subject", subject).Any("aud", aud).Msg("developer jwt sub and aud do not match")
+			}
+			return c.Next()
+		}
```
elffjs, line 63: Maybe swap these two sections?

The _subject_ won't in general be a dev license. I feel like the logic in that case is wrong?

internal/services/identity_service.go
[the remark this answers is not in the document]
elffjs: Move it to `valid_dev_license.go`, I think is the ask.

internal/middleware/valid_dev_license.go
```diff
@@ -0,0 +1,68 @@
...
+		), "*")
+
+		valid, err := idSvc.IsDevLicense(c.Context(), common.HexToAddress(subject))
+		if err != nil {
+			return err
+		}
+
+		if valid {
+			if !slices.Contains(aud, subject) {
+				logger.Debug().Str("subject", subject).Any("aud", aud).Msg("developer jwt sub and aud do not match")
```
elffjs, line 60: I'm surprised we don't error here?

internal/middleware/valid_dev_license.go
```diff
@@ -0,0 +1,68 @@
...
+		if err != nil {
+			return fiber.NewError(fiber.StatusBadRequest, err.Error())
+		}
+
+		subject := strings.Trim(string(
+			bytes.TrimSpace(
+				bytes.TrimSuffix(decoded, []byte("\x12\x04web3")),
+			),
+		), "*")
+
```
elffjs, line 52: I'm nervous about the logic from this point on. I think the pseudocode is like

```
if aud contains "dimo-driver":
  success
  
decode sub
if sub.conn_id != "web3":
  fail

addr = decode sub.user_id
if addr is not a dev license:
  fail

success

  
```

internal/services/identity_service.go
```diff
@@ -0,0 +1,124 @@
...
+func NewIdentityController(logger *zerolog.Logger, settings *config.Settings) *IdentityController {
+	return &IdentityController{
+		logger:      logger,
+		client:      &http.Client{},
+		identityURL: settings.IdentityURL,
+	}
+}
+
+const (
+	queryDevLicenseByClientId = `query ($clientId: Address!)
```
elffjs, line 36: I actually didn't know these could be anonymous.

elffjs, line 36: I agree with the linter here btw

internal/services/identity_service.go
```diff
@@ -0,0 +1,124 @@
...
+			"clientId": ethAddr.Hex(),
+		},
+	}
+
+	response, err := i.executeQuery(requestBody)
+	if err != nil {
+		return false, err
+	}
+
+	if len(response.Errors) > 1 {
```
elffjs, line 60: I think this is a good solution for now! I will make the dev license "404" response less terrible, I'll make an issue for that

elffjs, line 60: Let me make those Identity changes. It make a drastic difference whether it's a "404" or a "500", in my view.

internal/services/identity_service.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	req.Header.Set("Content-Type", "application/json")
+	resp, err := i.client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("graphql request failed making request: %v", err)
+	}
+	defer resp.Body.Close()
+
+	body, err := io.ReadAll(resp.Body)
```
elffjs, line 86: Good call to read this no matter what.

internal/services/identity_service.go
```diff
@@ -0,0 +1,124 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error reading response body: %v", err)
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("failed query response")
```
elffjs, line 92: As long as we're using `Errorf`, nice to put the status code in here.

internal/services/identity_service.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	if resp.StatusCode != http.StatusOK {
```
elffjs, line 95: Wait isn't this a repeat

charts/token-exchange-api/values-prod.yaml
```diff
@@ -13,6 +13,7 @@ env:
   USERS_API_GRPC_ADDRESS: users-api-prod:8086
   CONTRACT_ADDRESS_WHITELIST: 0xba5738a18d83d41847dffbdc6101d37c69c9b0cf,0x3b07e2A2ABdd0A9B8F7878bdE6487c502164B9dd
   CONTRACT_ADDRESS_SACD: '0x3c152B5d96769661008Ff404224d6530FCAC766d'
+  IDENTITY_URL: 'https://identity-api.dimo.zone/query'
```
elffjs, line 16: I worry we're going to get rate limited hard.

if we use the internal service name is it better? I think it will be better. In this case I think it would be 

```
http://identity-api-dev:8080/query
```

I think? Try that

cmd/token-exchange-api/api.go
KevinJoiner: We can't put this on the entire handler because then you would need an auth token to view `v1/swagger`.

internal/middleware/valid_dev_license.go
KevinJoiner: The `subject` should match the `ethereum_address` yes but in the example token provided from using LIWD the dev license is  in `aud` `0xf5ada890DA2E5582E38DF4648F9dAeE00e691199` and the vehicle owner is what is in subject. If you run this in identity you can see what I mean
```gql
 query Dev_licesnse{
  developerLicense(by:{clientId:"0xf5ada890DA2E5582E38DF4648F9dAeE00e691199"}){
    owner
    alias
    clientId
  }
  
  vehicles(first: 10, filterBy:{owner:"0x68D68B55b0fc11F4d0D91433A7F6217f2CCD02e3"})
{
  nodes{
    tokenId
  }
}    
}
```

internal/middleware/valid_dev_license.go
Allyson-English: I'm getting mixed messages about what we're checking for: https://github.com/DIMO-Network/token-exchange-api/pull/25#discussion_r1910403633 @KevinJoiner

elffjs: I got confused. The audience check is way up there. This is just hard to read.

My mistake. Disregard this and https://github.com/DIMO-Network/token-exchange-api/pull/25/files#r1909644540

internal/services/identity_service.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+type DeveloperLicense struct {
+	Owner    string         `json:"owner"`
+	Alias    string         `json:"alias"`
+	ClientId common.Address `json:"clientId"`
+}
+
+type Data struct {
+	DeveloperLicense DeveloperLicense `json:"developerLicense"`
```
elffjs, line 114: I think there's a decent chance this is null too.

internal/services/identity_service.go
```diff
@@ -0,0 +1,124 @@
...
+	if err := json.Unmarshal(body, &respBody); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal GraphQL response: %w", err)
+	}
+
+	return &respBody, nil
+}
+
+type DeveloperLicense struct {
+	Owner    string         `json:"owner"`
+	Alias    string         `json:"alias"`
```
elffjs, line 109: We do query for this, but are we doing anything with it?

internal/middleware/valid_dev_license.go
elffjs: I feel like there has to be a way to write this such that the control flow is more obvious. I'm afraid I don't hav a specific recommendation

internal/middleware/valid_dev_license.go
```diff
@@ -0,0 +1,68 @@
...
+			return fiber.NewError(fiber.StatusBadRequest, "failed to get audience from token claims")
+		}
+
+		if slices.Contains(aud, mobileAppAudience) {
+			return c.Next()
+		}
+
+		decoded, err := base64.RawStdEncoding.DecodeString(subj)
+		if err != nil {
+			return fiber.NewError(fiber.StatusBadRequest, err.Error())
```
elffjs, line 44: Similar to what you do above, I might avoid returning a raw Go error.

internal/middleware/valid_dev_license.go
```diff
@@ -0,0 +1,68 @@
...
+		}
+
+		subject := strings.Trim(string(
+			bytes.TrimSpace(
+				bytes.TrimSuffix(decoded, []byte("\x12\x04web3")),
+			),
+		), "*")
+
+		valid, err := idSvc.IsDevLicense(c.Context(), common.HexToAddress(subject))
```
elffjs, line 53: I think we should check that this thing is really an address. HexToAddress will accept any string more or less

internal/middleware/valid_dev_license.go
```diff
@@ -0,0 +1,68 @@
...
+
+		if valid {
+			if !slices.Contains(aud, subject) {
+				logger.Debug().Str("subject", subject).Any("aud", aud).Msg("developer jwt sub and aud do not match")
+			}
+			return c.Next()
+		}
+
+		logger.Debug().Msg("invalid dev license requesting access")
+		return fiber.NewError(fiber.StatusForbidden, fmt.Sprintf("invalid dev license requesting access: %s", subject))
```
elffjs, line 66: "invalid" is a broad term, to me. Can we just say it's not a dev license?

internal/services/identity_service.go
Allyson-English: we dont need any of them but I had to include something, these all just seemed like the might be relevant

elffjs: That's a good point. Can we at least make them match up? The struct and the query.

internal/middleware/valid_dev_license.go
KevinJoiner: Ignore this we don't want to accept this JWT

elffjs:
Here, much better:

image

internal/services/identity_service.go
```diff
@@ -51,9 +53,16 @@ func (i *IdentityController) IsDevLicense(ctx context.Context, ethAddr common.Ad
 	if err != nil {
 		return false, err
 	}
-	fmt.Println(response)
+
 	if len(response.Errors) >= 1 {
-		return false, nil
+		var errs []string
```
elffjs, line 58: I think we should make a hard distinction between "not found" and everything else.

elffjs: I'm okay to do this later this week/month but I do think we should do it.

internal/middleware/valid_dev_license.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/gofiber/fiber/v2"
+	"github.com/golang-jwt/jwt/v5"
+	"github.com/rs/zerolog"
+	"google.golang.org/protobuf/proto"
+)
+
+// mobileAppAudience Audience in DIMO mobile JWT
+const mobileAppAudience = "dimo-driver"
+
+//go:generate mockgen -source valid_dev_license.go -destination mocks/valid_dev_license_mock.go
```
KevinJoiner, line 20: nit: **This isn't important so you can ignore this rant of a comment***
Generally I don't think we gain much from creating mock packages for people to import. Instead we can define the go generate in the test file that needs it and give it an `_test.go` suffix. My reasoning is the current behavior creates the same type of problem when you define an interface in the implementing package. AKA If the token_exchange_test are updated to no longer need this mock we will still have this generated package hanging around unused. Secondly we are adding test specific code to our binary with little gain.
```suggestion
//go:generate mockgen -source ../middleware/valid_dev_license.go -destination mocks/valid_dev_license_mock_test.go
```
And if you made it this far thanks for reading this rant.

settings.sample.yaml
```diff
@@ -8,3 +8,4 @@ DEX_GRPC_ADDRESS: localhost:5557
 USERS_API_GRPC_ADDRESS: localhost:8089
 CONTRACT_ADDRESS_WHITELIST: 0x90c4d6113ec88dd4bdf12f26db2b3998fd13a144
 CONTRACT_ADDRESS_SACD: 0x123
+IDENTITY_URL: https://identity-api.dimo.zone/query
```
KevinJoiner, line 11: nit: you can add the feature flag here too.

KevinJoiner (approved):

internal/controllers/token_exchange_test.go
elffjs: @Allyson-English Why did we change this? Do you remember?

internal/services/identity_service.go
```diff
@@ -0,0 +1,129 @@
...
+	var respBody IdentityResponse
+	if err := json.Unmarshal(body, &respBody); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal GraphQL response: %v", err)
+	}
+
+	return &respBody, nil
+}
+
+type DeveloperLicense struct {
+	Alias string         `json:"alias"`
```
elffjs, line 114: If alias is null then is everything okay here? I think so but

internal/services/identity_service.go
```diff
@@ -0,0 +1,124 @@
...
+
+	req, err := http.NewRequest(http.MethodPost, i.identityURL, bytes.NewBuffer(jsonBody))
+	if err != nil {
+		return nil, fmt.Errorf("error creating request: %v", err)
+	}
+
+	req.Header.Set("Content-Type", "application/json")
+	resp, err := i.client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("graphql request failed making request: %v", err)
```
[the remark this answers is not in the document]
elffjs: Did we resolve this? I still see `%v`

internal/middleware/valid_dev_license.go
```diff
@@ -0,0 +1,77 @@
...
+		}
+
+		// get audience from claim to check if it is coming from dimo mobile
+		aud, err := token.Claims.GetAudience()
+		if err != nil {
+			return fiber.NewError(fiber.StatusBadRequest, "failed to get audience from token claims")
+		}
+
+		// no additional checks for mobile app
+		// TODO(ae): add additional security here eventually
```
elffjs, line 40: Sorry, what would that be?

internal/services/identity_service.go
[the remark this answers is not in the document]
elffjs: We have %v in a few other places too

elffjs (changes requested):
Some minor stuff

elffjs:
I'm okay to merge this and fix whatever I'm talking about later.

internal/middleware/valid_dev_license.go
Allyson-English: idk.. I put this here because this is what I understood the plan to be from you and Kevin based on the conversation you had with yev and others.

PR #29, opened 2025-03-25
WIP

LorranSutter:


internal/controllers/token_exchange.go
```diff
@@ -155,3 +218,152 @@ func (t *TokenExchangeController) GetDeviceCommandPermissionWithScope(c *fiber.C
...
+	// TODO Mayber a Regex is better. We should remove "ipfs://" from the SACD source
+	cid = strings.TrimPrefix(cid, "ipfs://")
+
+	url := fmt.Sprintf("https://assets.dimo.xyz//ipfs/%s", cid)
+
+	resp, err := http.Get(url)
+	if err != nil {
+		return "", fmt.Errorf("failed to fetch from IPFS: %w", err)
+	}
+	defer resp.Body.Close()
```
KevinJoiner, line 233: Couple of suggestions here
1. Make the Base URL configurable by adding it to settings then putting it on the controller struct in New(...)
2. Use [url.JoinPath](https://pkg.go.dev/net/url#JoinPath) Not only does it ensure you create a valid URL, But it also does character escaping so we are not vulnerable to XSS attacks when dealing with user provided data.
3. Create the request first with a context. If we do no then this request can hang indefinitely 
```suggestion
	ipfsURL := url.JoinPath(t.IPFSBaseURL, cid)
	req, err := http.NewRequestWithContext(ctx, http.MethodGet, ipfsURL, nil)
	if err != nil {
		return "", fmt.Errorf("failed to create HTTP request: %w", err)
	}
	resp, err := http.DefaultClient.Do(req)
	if err != nil {
		return "", fmt.Errorf("failed to execute request: %w", err)
	}
	defer resp.Body.Close()
```

internal/controllers/token_exchange.go
```diff
@@ -155,3 +218,152 @@ func (t *TokenExchangeController) GetDeviceCommandPermissionWithScope(c *fiber.C
...
+		mask.SetBit(mask, int(index*2), 1)
+		mask.SetBit(mask, int(index*2+1), 1)
+	}
+
+	return mask, nil
+}
+
+// TODO Documentation
+func (t *TokenExchangeController) validateAssetDID(did string, req *PermissionTokenRequest) (bool, error) {
+	decodedDID, err := DecodeNFTDID(did)
```
KevinJoiner, line 303:
```suggestion
	decodedDID, err := cloudevent.DecodeNFTDID(did)
```

internal/controllers/token_exchange.go
```diff
@@ -155,3 +218,152 @@ func (t *TokenExchangeController) GetDeviceCommandPermissionWithScope(c *fiber.C
...
+
+	// If we get here, the DID is valid for the given request
+	return true, nil
+}
+
+type DID struct {
+	Type            string         `json:"type"`
+	ChainID         uint64         `json:"chainId"`
+	ContractAddress common.Address `json:"contract"`
+	TokenID         uint32         `json:"tokenId"` // TODO Should it be big.int?
```
KevinJoiner, line 328: > // TODO Should it be big.int?

yeah let's update it next time we are in the cloudevent repo

internal/controllers/token_exchange.go
```diff
@@ -155,3 +218,152 @@ func (t *TokenExchangeController) GetDeviceCommandPermissionWithScope(c *fiber.C
...
+			continue
+		}
+
+		// Add permissions from this agreement
+		for _, permission := range agreement.Permissions {
+			userPermissions[permission.Name] = true
+		}
+	}
+
+	// TODO Check which permissions the user lacks
```
KevinJoiner, line 282: For now we can just expect the permission names in the JSON doc to be 1:1 with our privileges so something like this
```go
// privilege prefix to denote the 1:1 mapping to bit values and to make them easier to deprecate if desired in the future
var PermissionMap = map[string]int{
	"privilege:GetNonLocationHistory":   1, // All-time non-location data
	"privilege:ExecuteCommands":         2, // Commands
	"privilege:GetCurrentLocation":      3, // Current location
	"privilege:GetLocationHistory":      4, // All-time location
	"privilege:GetVinCredential":        5, // View VIN credential
	"privilege:GetLiveData":             6, // Subscribe live data
	"privilege:GetRawData":              7, // Raw data
	"privilege:GetApproximateLocation":  8, // Approximate location
}
```

KevinJoiner (changes requested):

internal/controllers/token_exchange.go
```diff
@@ -110,31 +126,78 @@ func (t *TokenExchangeController) GetDeviceCommandPermissionWithScope(c *fiber.C
...
+			t.logger.Warn().Err(err).Msg("Failed to validate IPFS JSON")
+		} else if hasPermFromSacdDoc {
+			return t.createAndReturnToken(c, pr, ethAddr)
 		}
+	}
 
-		if resMulti {
-			continue
+	// If the user doesn't have all permissions from IPFS doc, check bitstring
+	// Convert pr.Privileges to 2-bit array format
```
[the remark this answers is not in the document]
LorranSutter: I see. But the user can still set their permissions calling the SACD contract directly. We can't ensure they will make the doc correctly. Shouldn't we also consider that?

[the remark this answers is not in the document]
KevinJoiner: Yeah basically if the doc is made correctly aka it decodes with `type = dimo.sacd` Then that is the source of truth. If they have all bits set but no permission in a decoded doc we should not grant anything. Now if the doc does not decode we fallback to the bits.

elffjs:
@LorranSutter Can you rebase on main?

Is it fair to say that this is no longer WIP? Good to review?

LorranSutter:
Will solve these conflicts

internal/controllers/token_exchange.go
```diff
@@ -105,15 +147,258 @@ func (t *TokenExchangeController) GetDeviceCommandPermissionWithScope(c *fiber.C
...
+//   - error: An error describing why validation failed, or nil if validation succeeded
+func (t *TokenExchangeController) validateAssetDID(did string, req *PermissionTokenRequest) (bool, error) {
+	decodedDID, err := cloudevent.DecodeNFTDID(did)
+	if err != nil {
+		return false, fmt.Errorf("failed to decode DID: %w", err)
+	}
+
+	requestNFTAddr := common.HexToAddress(req.NFTContractAddress)
+
+	if decodedDID.ContractAddress != requestNFTAddr {
```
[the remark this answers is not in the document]
elffjs: `common.Address`, assuming that's what this is, is `[20]byte`. So checking equality would be fine. There's also a `Cmp` method.

internal/controllers/token_exchange.go
```diff
@@ -105,15 +147,258 @@ func (t *TokenExchangeController) GetDeviceCommandPermissionWithScope(c *fiber.C
...
+	pr *PermissionTokenRequest,
+	ethAddr *common.Address,
+) error {
+	// Convert pr.Privileges to 2-bit array format
+	mask, err := intArrayTo2BitArray(pr.Privileges, 128) // Assuming max privilege is 128
+	if err != nil {
+		return fiber.NewError(fiber.StatusBadRequest, err.Error())
 	}
 
 	ret, err := s.GetPermissions(nil, nftAddr, big.NewInt(pr.TokenID), *ethAddr, mask)
```
elffjs, line 404: suggestion: Does the `resPermRecord` you get before this have the bit string and the expiration? Can we skip this call?

internal/controllers/token_exchange.go
```diff
@@ -105,15 +147,258 @@ func (t *TokenExchangeController) GetDeviceCommandPermissionWithScope(c *fiber.C
...
+// and verifies that it has the correct type for a DIMO SACD document.
+//
+// Parameters:
+//   - ctx: The context for the IPFS request, which can be used for cancellation and timeouts
+//   - source: The IPFS content identifier (CID) for the SACD document, typically with an "ipfs://" prefix
+//
+// Returns:
+//   - *PermissionRecord: A pointer to the parsed permission record if valid, or nil if the document
+//     could not be fetched, parsed, or doesn't have the correct type
+func (t *TokenExchangeController) hasValidSacdDoc(ctx context.Context, source string) *PermissionRecord {
```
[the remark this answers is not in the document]
elffjs: suggestion: To me, a function named `has_` usually returns a boolean. The return type here is surprising to me.

internal/controllers/token_exchange.go
```diff
@@ -105,15 +147,258 @@ func (t *TokenExchangeController) GetDeviceCommandPermissionWithScope(c *fiber.C
...
+	if err != nil {
+		t.logger.Warn().Err(err).Msg("Failed to fetch JSON from IPFS")
+		return nil
+	}
+
+	var record PermissionRecord
+	if err := json.Unmarshal(sacdDoc, &record); err != nil {
+		t.logger.Warn().Err(err).Msg(fmt.Sprintf("invalid JSON format: %v", err))
+		return nil
+	}
```
elffjs, line 209: note: I worry here about distinguishing between network/IPFS errors and "they didn't give a real URL" and "the document is there, but it's invalid".

internal/controllers/token_exchange.go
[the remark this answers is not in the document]
elffjs: Let's return errors here, yeah.

internal/controllers/token_exchange.go
```diff
@@ -105,15 +147,258 @@ func (t *TokenExchangeController) GetDeviceCommandPermissionWithScope(c *fiber.C
...
+//
+// Returns:
+//   - []byte: The content retrieved from IPFS as a byte slice
+//   - error: An error if the request fails at any stage (URL construction, HTTP request creation,
+//     request execution, or response reading)
+func (t *TokenExchangeController) fetchFromIPFS(ctx context.Context, cid string) ([]byte, error) {
+	cid = strings.TrimPrefix(cid, "ipfs://")
+
+	ipfsURL, err := url.JoinPath(t.settings.IPFSBaseURL, cid)
+	if err != nil {
```
elffjs, line 235: suggestion: You can get rid of the potential for error here by parsing `IPFSBaseURL` once, at the start of the program.

internal/controllers/token_exchange.go
```diff
@@ -105,15 +147,258 @@ func (t *TokenExchangeController) GetDeviceCommandPermissionWithScope(c *fiber.C
...
+		return nil, fmt.Errorf("failed to join URL paths: %w", err)
+	}
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, ipfsURL, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create HTTP request: %w", err)
+	}
+
+	resp, err := http.DefaultClient.Do(req)
```
elffjs, line 244: thought: I wonder if we can gauge the file's size before downloading. Maybe someone makes us download some heinously large file.

internal/controllers/token_exchange.go
```diff
@@ -105,15 +147,258 @@ func (t *TokenExchangeController) GetDeviceCommandPermissionWithScope(c *fiber.C
...
+// Returns:
+//   - error: An error if the document is invalid, expired, or missing requested permissions;
+//     nil if all permissions are valid and the token is successfully created and returned
+func (t *TokenExchangeController) evaluateSacdDoc(c *fiber.Ctx, record PermissionRecord, pr *PermissionTokenRequest, grantee *common.Address) error {
+	now := time.Now()
+	if now.Before(record.Data.EffectiveAt) || now.After(record.Data.ExpiresAt) {
+		return fiber.NewError(fiber.StatusBadRequest, "Permission record is expired or not yet effective")
+	}
+
+	if record.Data.Grantee.Address != grantee.Hex() {
```
elffjs, line 277: Maybe we should parse `Grantee` and so on into an `Address` so we don't have to worry about case here?

internal/controllers/token_exchange.go
```diff
@@ -105,15 +147,258 @@ func (t *TokenExchangeController) GetDeviceCommandPermissionWithScope(c *fiber.C
...
+//   - record: The SACD permission record containing the granted permissions and validity period
+//   - pr: The permission token request containing the requested privileges and token information
+//   - grantee: The Ethereum address of the user requesting permissions
+//
+// Returns:
+//   - error: An error if the document is invalid, expired, or missing requested permissions;
+//     nil if all permissions are valid and the token is successfully created and returned
+func (t *TokenExchangeController) evaluateSacdDoc(c *fiber.Ctx, record PermissionRecord, pr *PermissionTokenRequest, grantee *common.Address) error {
+	now := time.Now()
+	if now.Before(record.Data.EffectiveAt) || now.After(record.Data.ExpiresAt) {
```
elffjs, line 273: Do the individual agreements have times we need to worry about as well?

elffjs: Is either one of these optional?

internal/controllers/token_exchange.go
```diff
@@ -105,15 +147,258 @@ func (t *TokenExchangeController) GetDeviceCommandPermissionWithScope(c *fiber.C
...
+//
+// Parameters:
+//   - did: The decentralized identifier string to validate, typically in the format "did:nft:..."
+//   - req: The permission token request containing the NFT contract address and token ID to match against
+//
+// Returns:
+//   - bool: true if the DID is valid and matches the request parameters, false otherwise
+//   - error: An error describing why validation failed, or nil if validation succeeded
+func (t *TokenExchangeController) validateAssetDID(did string, req *PermissionTokenRequest) (bool, error) {
+	decodedDID, err := cloudevent.DecodeNFTDID(did)
```
elffjs, line 355: Are we worried about chain id here?

internal/controllers/token_exchange.go
```diff
@@ -105,15 +147,258 @@ func (t *TokenExchangeController) GetDeviceCommandPermissionWithScope(c *fiber.C
...
+//   - c: The Fiber context for the HTTP request
+//   - s: The SACD contract instance used to check permissions
+//   - nftAddr: The Ethereum address of the NFT contract
+//   - pr: The permission token request containing token ID and requested privileges
+//   - ethAddr: The Ethereum address of the user requesting permissions
+//
+// Returns:
+//   - error: An error if the user lacks any requested permissions or if there's a system error,
+//     otherwise nil if the token is successfully created and returned
+func (t *TokenExchangeController) evaluatePermissionsBits(
```
elffjs, line 391: nitpick: Massive argument list.

Something we can think about later.

internal/controllers/token_exchange.go
```diff
@@ -105,15 +147,258 @@ func (t *TokenExchangeController) GetDeviceCommandPermissionWithScope(c *fiber.C
...
+			// If we don't have a mapping for this privilege ID, consider it missing
+			missingPermissions = append(missingPermissions, privID)
+			continue
+		}
+
+		// Check if the user has this permission
+		if !userPermissions[permName] {
+			missingPermissions = append(missingPermissions, privID)
+		}
+	}
```
elffjs, line 317: note: This seems to work.

internal/controllers/token_exchange.go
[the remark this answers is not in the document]
KevinJoiner: ah I forget it's an array not a slice

internal/controllers/token_exchange.go
LorranSutter: It does, but `GetPermissions` also checks if the `ethAddr` is the token owner

internal/controllers/token_exchange.go
```diff
@@ -145,28 +430,10 @@ func (t *TokenExchangeController) GetDeviceCommandPermissionWithScope(c *fiber.C
 			}
```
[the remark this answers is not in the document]
LorranSutter: I don't think so. We can do it in another PR. This is already too big

internal/contracts/contracts.go
```diff
@@ -50,5 +50,6 @@ type MultiPriv interface {
 }
 
 type Sacd interface {
+	CurrentPermissionRecord(opts *bind.CallOpts, asset common.Address, tokenID *big.Int, grantee common.Address) (sacd.ISacdPermissionRecord, error)
 	GetPermissions(opts *bind.CallOpts, asset common.Address, tokenID *big.Int, grantee common.Address, permissions *big.Int) (*big.Int, error)
```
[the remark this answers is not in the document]
LorranSutter: To check if the `grantee` is the owner of the tokenID

internal/controllers/token_exchange.go
LorranSutter: The OG SACD does not define times for individual agreements, and I believe we stuck to that. But it bothers me.

I don't think they are optional

internal/controllers/token_exchange.go
```diff
@@ -42,6 +62,28 @@ type PermissionTokenRequest struct {
 type PermissionTokenResponse struct {
 	Token string `json:"token"`
 }
+type PermissionRecord struct {
```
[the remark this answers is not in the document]
LorranSutter: You mean this? https://github.com/DIMO-Network/cloudevent/blob/main/cloudevent.go#L31
There are so many other elements

internal/controllers/token_exchange.go
KevinJoiner: yeah I think we will want this in the future

internal/controllers/token_exchange_test.go
```diff
@@ -50,13 +51,23 @@ func TestTokenExchangeController_GetDeviceCommandPermissionWithScope(t *testing.
...
 	// setup app and route req
-	c := NewTokenExchangeController(&logger, &config.Settings{
+	c, err := NewTokenExchangeController(&logger, &config.Settings{
 		BlockchainNodeURL:        "http://testurl.com/mock",
 		ContractAddressWhitelist: "",
 		ContractAddressSacd:      "0xa6",
 	}, dexService, usersSvc, contractsMgr, &client)
+	if err != nil {
+		logger.Fatal().Err(err).Msg("Failed to initialize token exchange controller")
+	}
```
KevinJoiner, line 61:
```suggestion
	require.NoError(t, err,"Failed to initialize token exchange controller")
```

internal/models/permission_record.go
```diff
@@ -0,0 +1,29 @@
...
+		ExpiresAt   time.Time `json:"expiresAt"`
+		Agreements  []struct {
+			Type        string `json:"type"`
+			Asset       string `json:"asset"`
+			Permissions []struct {
+				Name string `json:"name"`
+			} `json:"permissions"`
+		} `json:"agreements"`
+	} `json:"data"`
+}
```
KevinJoiner, line 29: Break this up to something like this so we are not using nested structs
```suggestion
// PermissionRecord is the top-level structure for permission events
type PermissionRecord struct {
	SpecVersion string         `json:"specversion"`
	Timestamp   time.Time      `json:"timestamp"`
	Type        string         `json:"type"`
	Data        PermissionData `json:"data"`
}

// PermissionData contains the core permission data
type PermissionData struct {
	Grantor     Address     `json:"grantor"`
	Grantee     Address     `json:"grantee"`
	EffectiveAt time.Time   `json:"effectiveAt"`
	ExpiresAt   time.Time   `json:"expiresAt"`
	Agreements  []Agreement `json:"agreements"`
}

// Agreement represents a permission agreement for an asset
type Agreement struct {
	Type        string       `json:"type"`
	Asset       string       `json:"asset"`
	Permissions []Permission `json:"permissions"`
}

// Permission defines a single permission
type Permission struct {
	Name string `json:"name"`
}

// Address represents a blockchain address
type Address struct {
	Address string `json:"address"`
}
```

KevinJoiner (changes requested):
couple small things

internal/controllers/token_exchange.go
```diff
@@ -105,15 +137,253 @@ func (t *TokenExchangeController) GetDeviceCommandPermissionWithScope(c *fiber.C
...
+// and verifies that it has the correct type for a DIMO SACD document.
+//
+// Parameters:
+//   - ctx: The context for the IPFS request, which can be used for cancellation and timeouts
+//   - source: The IPFS content identifier (CID) for the SACD document, typically with an "ipfs://" prefix
+//
+// Returns:
+//   - *PermissionRecord: A pointer to the parsed permission record if valid, or nil if the document
+//     could not be fetched, parsed, or doesn't have the correct type
+func (t *TokenExchangeController) getValidSacdDoc(ctx context.Context, source string) (*models.PermissionRecord, error) {
```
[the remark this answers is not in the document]
LorranSutter: This is checking this field https://github.com/DIMO-Network/sacd/blob/main/data/sacd.template.json#L4
An attestation would still be a sacd, no? And we would check if it is an attestation here https://github.com/DIMO-Network/sacd/blob/main/data/sacd.template.json#L23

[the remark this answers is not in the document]
Allyson-English: ah, yep-- you're right!

KevinJoiner:
you will need to merge main into this so we can run CI

internal/controllers/token_exchange_test.go
```diff
@@ -50,13 +51,23 @@ func TestTokenExchangeController_GetDeviceCommandPermissionWithScope(t *testing.
...
 	// setup app and route req
-	c := NewTokenExchangeController(&logger, &config.Settings{
+	c, err := NewTokenExchangeController(&logger, &config.Settings{
 		BlockchainNodeURL:        "http://testurl.com/mock",
 		ContractAddressWhitelist: "",
 		ContractAddressSacd:      "0xa6",
 	}, dexService, usersSvc, contractsMgr, &client)
+	if err != nil {
+		require.NoError(t, err, "Failed to initialize token exchange controller")
+	}
```
KevinJoiner, line 61: nit: you don't need the nil check either but this works either way
```suggestion
	require.NoError(t, err, "Failed to initialize token exchange controller")
```

internal/controllers/token_exchange.go
KevinJoiner: maybe a TODO so we get to it later

KevinJoiner (approved):
Excited to give this a try.

internal/contracts/contracts.go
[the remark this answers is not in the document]
elffjs: This seems like a very roundabout way of doing that.

internal/models/permission_record.go
```diff
@@ -0,0 +1,39 @@
...
+
+// Permission defines a single permission
+type Permission struct {
+	Name string `json:"name"`
+}
+
+// Address represents a blockchain address
+type Address struct {
+	Address string `json:"address"`
```
elffjs, line 38: Could use `common.Address` here. There are tradeoffs.

elffjs (approved):
All good

internal/controllers/token_exchange.go
```diff
@@ -105,15 +132,253 @@ func (t *TokenExchangeController) GetDeviceCommandPermissionWithScope(c *fiber.C
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch JSON from IPFS: %w", err)
+	}
+
+	var record models.PermissionRecord
+	if err := json.Unmarshal(sacdDoc, &record); err != nil {
+		return nil, fmt.Errorf("invalid JSON format: %w", err)
+	}
+
+	if record.Type != "dimo.sacd" {
```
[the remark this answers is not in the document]
elffjs: Never mind. I think this is fine.

elffjs:
Merge it?

PR #30, opened 2025-03-31
Quick exch

elffjs:


internal/contracts/sacd/sacd.json
```diff
@@ -0,0 +1,36 @@
+[
```
elffjs, line 1: Need to make this a `go:generate` operation.

PR #32, opened 2025-04-22
ex-3544-sacd-extended-attestation-permissions

Allyson-English:
As I'm working on this PR, I want to revisit a prior decision about the contents of the attestation agreement (effectiveAt/ expiresAt):

```javascript
{
         "type": "cloudevent", # required
         "eventType": "dimo.attestation", # required 
	 "source": "0x3330000000000000000000000000000000000000", # optional
	 "ids": ["attestation_1", "attestation_2", "attestation_3"], # optional
         "effectiveAt": "2025-04-16T15:12:00Z", # optional 
         "expiresAt": "2030-12-20T05:20:45Z", # optional
}

* in the case of conflicting dates, 
  the optional effectiveAt and expiresAt fields are superceeded 
  by whatever information is (1) on chain and (2) found in the parent section of the doc
```

if we're going to overwrite whatever is inside the agreement with (1) whatever is on chain and (2) whatever we find in the parent doc, then it doesn't seem like there's a need to be checking this here, or include it at all. 

curious what others think about removing this or what people think is the benefit to using there (is there a case where this would help the user rather than just make things more confusing for them)


internal/controllers/token_exchange.go
```diff
@@ -285,13 +259,56 @@ func (t *TokenExchangeController) evaluateSacdDoc(c *fiber.Ctx, record *models.P
...
+				missing = append(missing, attID)
+			}
+		}
+
+		if len(missing) >= 1 {
+			errs = append(errs, fmt.Errorf("for source %s missing grants for attestation ids: %s", claim.Source, strings.Join(missing, ", ")))
+		}
+	}
+
+	return errors.Join(errs...)
```
KevinJoiner, line 311: You don't need a list for this
```suggestion
	var errs error
	for _, claim := range tokenReq.Attestations {
		att, ok := attestations[claim.Source]
		if !ok {
			errs = append(errs, fmt.Errorf("missing grant for source: %s", claim.Source))
			continue
		}

		var missing []string
		for _, attID := range claim.AttestationIDs {
			_, ok := att[attID]

			if !ok {
				missing = append(missing, attID)
			}
		}

		if len(missing) >= 1 {
			err = errors.Join(fmt.Errorf("for source %s missing grants for attestation ids: %s", claim.Source, strings.Join(missing, ", "))
		}
	}

	return errs
```

elffjs:
I'm so out of it, I thought this was in already.

Appreciate the PR description. Taking a look.

elffjs:
Looks like we need to generate something?

elffjs:
Looks like Swagger and mocks. Maybe my fault, let me see if they're failing on main.

elffjs:
Still one job failing

elffjs:
Did we come to a decision on the question in the original comment here?

internal/services/ipfs_service.go
```diff
@@ -0,0 +1,63 @@
...
+		return nil, fmt.Errorf("failed to create HTTP request: %w", err)
+	}
+
+	resp, err := i.client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("failed to execute request: %w", err)
+	}
+	defer resp.Body.Close() //nolint:errcheck
+
+	return io.ReadAll(resp.Body)
```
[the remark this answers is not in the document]
elffjs: note: I like the error wrapping elsewhere in this function.

internal/services/ipfs_service.go
```diff
@@ -0,0 +1,63 @@
...
+// Returns:
+//   - []byte: The content retrieved from IPFS as a byte slice
+//   - error: An error if the request fails at any stage (URL construction, HTTP request creation,
+//     request execution, or response reading)
+func (i *IPFSController) FetchFromIPFS(ctx context.Context, cid string) ([]byte, error) {
+	cid = strings.TrimPrefix(cid, "ipfs://")
+
+	ipfsURL := i.ipfsBaseURL.JoinPath(cid).String()
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, ipfsURL, nil)
```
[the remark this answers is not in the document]
elffjs: note: Good context use.

internal/controllers/token_exchange.go
```diff
@@ -169,7 +170,7 @@ func (t *TokenExchangeController) createAndReturnToken(c *fiber.Ctx, pr *Permiss
 //   - *PermissionRecord: A pointer to the parsed permission record if valid, or nil if the document
 //     could not be fetched, parsed, or doesn't have the correct type
 func (t *TokenExchangeController) getValidSacdDoc(ctx context.Context, source string) (*models.PermissionRecord, error) {
-	sacdDoc, err := t.fetchFromIPFS(ctx, source)
+	sacdDoc, err := t.ipfsService.FetchFromIPFS(ctx, source)
```
elffjs, line 173: nitpick: There’s a bit of stutter here with the variable and method names.

internal/controllers/token_exchange.go
```diff
@@ -285,13 +265,92 @@ func (t *TokenExchangeController) evaluateSacdDoc(c *fiber.Ctx, record *models.P
...
+	return nil
+}
+
+func (t *TokenExchangeController) evaluateAttestations(record *models.PermissionRecord, tokenReq *PermissionTokenRequest) error {
+	if valid, err := t.validateAssetDID(record.Data.Asset, tokenReq); err != nil || !valid {
+		return fmt.Errorf("failed to validate attestation asset: %s", record.Data.Asset)
+	}
+
+	if time.Now().Before(record.Data.EffectiveAt) {
+		return fmt.Errorf("agreement inactive, effective starting at: %s", record.Data.EffectiveAt.String())
```
elffjs, line 280: note: I wonder if the caller will have a hard time figuring out which agreement we’re talking about in these messages.

internal/controllers/token_exchange.go
```diff
@@ -285,13 +265,92 @@ func (t *TokenExchangeController) evaluateSacdDoc(c *fiber.Ctx, record *models.P
...
+	}
+
+	if record.Data.ExpiresAt.Before(time.Now()) {
+		return fmt.Errorf("agreement inactive, expired at: %s", record.Data.ExpiresAt.String())
+	}
+
+	var allAttestations bool
+	attestations := make(map[string]map[string]struct{})
+	for _, agreement := range record.Data.Agreements {
+		if agreement.Type != "cloudevent" || agreement.EventType != cloudevent.TypeAttestation {
```
elffjs, line 290: Could someone explain this `type` thing to me?

internal/controllers/token_exchange.go
```diff
@@ -285,13 +265,92 @@ func (t *TokenExchangeController) evaluateSacdDoc(c *fiber.Ctx, record *models.P
...
+
+	if time.Now().Before(record.Data.EffectiveAt) {
+		return fmt.Errorf("agreement inactive, effective starting at: %s", record.Data.EffectiveAt.String())
+	}
+
+	if record.Data.ExpiresAt.Before(time.Now()) {
+		return fmt.Errorf("agreement inactive, expired at: %s", record.Data.ExpiresAt.String())
+	}
+
+	var allAttestations bool
```
elffjs, line 287: nitpick: Maybe explicitly set this to false? Not a big deal.

internal/controllers/token_exchange.go
```diff
@@ -285,13 +265,92 @@ func (t *TokenExchangeController) evaluateSacdDoc(c *fiber.Ctx, record *models.P
...
+		if agreement.Type != "cloudevent" || agreement.EventType != cloudevent.TypeAttestation {
+			return errors.New("unexpected types in attestation agreement")
+		}
+
+		if agreement.Source == nil {
+			allAttestations = true
+			continue
+		}
+
+		if agreement.EffectiveAt != nil && !agreement.EffectiveAt.IsZero() {
```
elffjs, line 299: note: Interesting handling the zero case.

internal/controllers/token_exchange.go
```diff
@@ -285,13 +265,92 @@ func (t *TokenExchangeController) evaluateSacdDoc(c *fiber.Ctx, record *models.P
...
+	if time.Now().Before(record.Data.EffectiveAt) {
+		return fmt.Errorf("agreement inactive, effective starting at: %s", record.Data.EffectiveAt.String())
+	}
+
+	if record.Data.ExpiresAt.Before(time.Now()) {
+		return fmt.Errorf("agreement inactive, expired at: %s", record.Data.ExpiresAt.String())
+	}
+
+	var allAttestations bool
+	attestations := make(map[string]map[string]struct{})
```
elffjs, line 288: nitpick: It might be good to comment on what the strings are.

internal/controllers/token_exchange.go
```diff
@@ -245,6 +210,21 @@ func (t *TokenExchangeController) evaluateSacdDoc(c *fiber.Ctx, record *models.P
...
+		return fiber.NewError(fiber.StatusBadRequest, "failed to validate requested permissions")
+	}
+
+	if err := t.evaluateAttestations(record, pr); err != nil {
+		t.logger.Err(err).Str("grantee", grantee.Hex()).Msg("failed to validate requested attestations")
+		return fiber.NewError(fiber.StatusBadRequest, "failed to validate requested permissions")
+	}
+
+	// If we get here, all permissions are valid
+	return t.createAndReturnToken(c, pr, grantee)
```
elffjs, line 224: blocking: Are we modifying the token at all in this PR? It’s OK if we’re not, I just want to know.

internal/controllers/token_exchange.go
```diff
@@ -285,13 +265,92 @@ func (t *TokenExchangeController) evaluateSacdDoc(c *fiber.Ctx, record *models.P
...
+	var errs error
+	for _, claim := range tokenReq.Attestations {
+		// if no source specified,
+		// user can view all asset attestations
+		if claim.Source == nil {
+			errs = errors.Join(errs, errors.New("requesting access to all attestations but granted limited set"))
+			continue
+		}
+
+		// if we get there the soure should be included
```
elffjs, line 332:
```suggestion
		// if we get there the source should be included
```

internal/controllers/token_exchange.go
```diff
@@ -285,13 +259,56 @@ func (t *TokenExchangeController) evaluateSacdDoc(c *fiber.Ctx, record *models.P
...
+	var errs []error
+	for _, claim := range tokenReq.Attestations {
+		att, ok := attestations[claim.Source]
+		if !ok {
+			errs = append(errs, fmt.Errorf("missing grant for source: %s", claim.Source))
+			continue
+		}
+
+		var missing []string
+		for _, attID := range claim.AttestationIDs {
```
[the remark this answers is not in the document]
elffjs: Agree with all this.

internal/controllers/token_exchange.go
elffjs: Feel free to pull in a set type if it helps readability.

elffjs: note: This is an area where I would write honest to God unit tests. I will check the test file in a second.

Allyson-English:
> Did we come to a decision on the question in the original comment here?

yes--we confirmed that we want to check both and that the child effective/ expires at dates should be locked to the parent... I ended up doing this by first checking the parent dates are valid for the current time and then check if the child dates (if they exist) are valid

this would not result in an error if the child dates were valid but had an expiration that was later than the parent expiration. I think, however, that's probably fine because the token exchange isn't the place to check that (imo)

internal/controllers/token_exchange.go
```diff
@@ -285,13 +265,92 @@ func (t *TokenExchangeController) evaluateSacdDoc(c *fiber.Ctx, record *models.P
...
+	}
+
+	var allAttestations bool
+	attestations := make(map[string]map[string]struct{})
+	for _, agreement := range record.Data.Agreements {
+		if agreement.Type != "cloudevent" || agreement.EventType != cloudevent.TypeAttestation {
+			return errors.New("unexpected types in attestation agreement")
+		}
+
+		if agreement.Source == nil {
```
elffjs, line 294: blocking: Not specifying a source but specifying some attestation ids: do we

1. Ignore the ids (as we seem to do here)
2. Respect the ids (awkward, since ids aren't unique across sources)
3. Throw an error

What do you guys think? I just want to see it in writing.

internal/controllers/token_exchange.go
elffjs: blocking: Are we going to throw an error when this loop examines one of the previous "permission" agreements? [The example.](https://github.com/DIMO-Network/sacd/blob/main/data/sacd.permission.example.json)

internal/controllers/token_exchange.go
Allyson-English: I think respecting the ids could be interesting/ potentially valuable (maybe there's a standard id used for the type of vehicle check up someone needs to get from a mechanic before they register their car)..  open to others thoughts as well!

internal/controllers/token_exchange.go
Allyson-English: I added an attestations field to the token but I'm not modifying in the sense that it changes at all while we're validating.. we're just comparing what's included in it to what's in the sacd

internal/controllers/token_exchange.go
```diff
@@ -104,8 +104,8 @@ func (t *TokenExchangeController) GetDeviceCommandPermissionWithScope(c *fiber.C
 
 	t.logger.Debug().Interface("request", pr).Msg("Got request.")
 
-	if len(pr.Privileges) == 0 {
-		return fiber.NewError(fiber.StatusBadRequest, "Please provide at least one privilege.")
+	if len(pr.Privileges) == 0 && len(pr.Attestations) == 0 {
+		return fiber.NewError(fiber.StatusBadRequest, "Please provide at least one privilege or attestation claim.")
```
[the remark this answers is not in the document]
Allyson-English: is there a reason why someone would be using token exchange in that case? I agree I think it'd be fine but would want it to be consistent what happens if someone requests no privileges

internal/controllers/token_exchange.go
Allyson-English: the thinking with `type` was that eventually we might want a way for someone to share other types of cloudevents (such as `status`) 

regarding your second comment--, yes as written it would. what do you think about [this solution](https://github.com/DIMO-Network/token-exchange-api/pull/32/commits/b2556b328aa2684232bfef8e844dad1a2478be4e)

internal/models/permission_record.go
```diff
@@ -18,12 +18,18 @@ type PermissionData struct {
...
 	EffectiveAt time.Time   `json:"effectiveAt"`
 	ExpiresAt   time.Time   `json:"expiresAt"`
+	Asset       string      `json:"asset,omitempty"`
 	Agreements  []Agreement `json:"agreements"`
 }
 
 // Agreement represents a permission agreement for an asset
 type Agreement struct {
 	Type        string       `json:"type"`
+	EventType   string       `json:"eventType"`
```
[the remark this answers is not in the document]
elffjs: Oh, I thought this was in total contrast to the `Attest`.

internal/controllers/token_exchange.go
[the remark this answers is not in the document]
elffjs: I think the way it is is fine. I sort of suspect that some people want to use token-exchange like a permission checker.

internal/controllers/token_exchange.go
elffjs: I like the look of it.

internal/controllers/token_exchange.go
elffjs: It sounds like we don't want to do (2). (1) or (3)?

internal/services/ipfs_service.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+func NewIPFSClient(logger *zerolog.Logger, settings *config.Settings) (*IPFSClient, error) {
+	ipfsBaseURL, err := url.Parse(settings.IPFSBaseURL)
+	if err != nil {
+		return nil, fmt.Errorf("invalid IPFS base URL: %w", err)
+	}
+
+	if settings.IPFSTimeout == 0 {
+		settings.IPFSTimeout = DefaultIPFSTimeout
```
elffjs, line 32: note: Nice touch.

internal/controllers/token_exchange.go
elffjs: blocking: I think the attestations are in the request, and I agree they're in the "DTO", but I don't think we're sending them to Dex and I don't think they're ending up in the final token we return to the user.

See [this method](https://github.com/DIMO-Network/shared/blob/1a95c46a8ef25156ed6ef42b12f2a0dd4e4ebcdc/pkg/middleware/privilegetoken/claims.go#L26). If you want to punt on this and to it in another PR, I'm totally fine.

I have to apologize for how this is organized.

elffjs: Probably we should pull that stuff out of shared and back into here. What does everyone think? @KevinJoiner @LorranSutter

internal/controllers/token_exchange.go
```diff
@@ -58,25 +62,22 @@ type PermissionTokenRequest struct {
 	NFTContractAddress string `json:"nftContractAddress" example:"0xbA5738a18d83D41847dfFbDC6101d37C69c9B0cF" validate:"required"`
 	// Audience is the intended audience for the token.
 	Audience []string `json:"audience" validate:"optional"`
+	// Attestations
+	Attestations []services.Attestation `json:"attestations"`
```
elffjs, line 66: blocking: Do we actually want to expose all of the fields of `services.Attestation`? `EffectiveAt` and `ExpiresAt` seem awkward.

It seems like we're also expecting these to be in the resulting JWT. It's not clear to me how a downstream service like Fetch should interpret these.

internal/controllers/token_exchange.go
```diff
@@ -245,6 +211,26 @@ func (t *TokenExchangeController) evaluateSacdDoc(c *fiber.Ctx, record *models.P
 		return fiber.NewError(fiber.StatusBadRequest, "Grantee address in permission record doesn't match requester")
 	}
 
+	for _, agg := range record.Data.Agreements {
+		switch agg.Type {
+		case "cloudevent":
+			if err := t.evaluateAttestations(record, pr); err != nil {
```
elffjs, line 217: blocking: It feels like we should be passing in `agg` somehow.

internal/controllers/token_exchange.go
elffjs: Nudging this. I'll read the code too but curious what your thought is.

internal/services/dex_service.go
```diff
@@ -30,6 +30,13 @@ type PrivilegeTokenDTO struct {
 	PrivilegeIDs       []int64
 	NFTContractAddress string
 	Audience           []string
+	Attestations       []Attestation
+}
+
+type Attestation struct {
+	EventType      string   `json:"eventType"`
+	Source         *string  `json:"source"`
+	AttestationIDs []string `json:"id"`
```
elffjs, line 39: note: A bit strange that it's `id` and not `ids`.

internal/models/permission_record.go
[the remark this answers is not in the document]
elffjs: I still think there's something funky here. Not hurting anyone though.

internal/controllers/token_exchange.go
```diff
@@ -245,26 +211,96 @@ func (t *TokenExchangeController) evaluateSacdDoc(c *fiber.Ctx, record *models.P
...
+				return fiber.NewError(fiber.StatusBadRequest, FailedToValidateRequest)
+			}
+		}
+	}
+
+	// If we get here, all permission and attestation claims are valid
+	return t.createAndReturnToken(c, tokenReq, grantee)
+}
+
+var AllAttestations = "ALL"
```
elffjs, line 270: This scares me a bit.

internal/services/dex_service.go
elffjs: Oh but I see `ids` in the example here 

https://github.com/DIMO-Network/sacd/blob/3336a345958ea22a86b7559c345323ba4a49fbe0/data/sacd.attestation.example.json#L23

Which is it?

internal/controllers/token_exchange.go
```diff
@@ -126,7 +127,7 @@ func (t *TokenExchangeController) GetDeviceCommandPermissionWithScope(c *fiber.C
 
 	record, err := t.getValidSacdDoc(c.Context(), resPermRecord.Source)
 	if err != nil {
-		t.logger.Warn().Err(err).Msg("Failed to get valid SACD document")
+		t.logger.Warn().Err(err).Msg("failed to get valid SACD document")
 		// If the user doesn't have a valid IPFS doc, check bitstring
 		return t.evaluatePermissionsBits(c, s, nftAddr, pr, ethAddr)
```
elffjs, line 132: blocking: Isn't this dangerous? It seems like we'll let people view any attestation if the bits (which don't deal with attestations) are right?

I think if all of the following are true:

* You're asking to view attestations
* You're not the vehicle owner
* We can't find a SACD doc

Then we ought to error.

elffjs: Okay to push this to another PR if you want, but I think we do have to handle it.

internal/controllers/token_exchange.go
```diff
@@ -245,26 +211,96 @@ func (t *TokenExchangeController) evaluateSacdDoc(c *fiber.Ctx, record *models.P
...
+			if agreement.Source == nil {
+				source = &AllAttestations
+			}
 
-		// Validate the asset DID if it exists in the record
-		valid, err := t.validateAssetDID(agreement.Asset, pr)
-		if err != nil || !valid {
-			continue
-		}
+			attestations[*source] = shared.NewStringSet()
```
elffjs, line 286: We might overwrite something here, no? If you had multiple grants that overlapped in the source.

internal/controllers/token_exchange.go
```diff
@@ -245,26 +211,96 @@ func (t *TokenExchangeController) evaluateSacdDoc(c *fiber.Ctx, record *models.P
...
+
+			if agg.EffectiveAt != nil && !agg.EffectiveAt.IsZero() {
+				if time.Now().Before(*agg.EffectiveAt) {
+					t.logger.Err(err).Str("grantee", grantee.Hex()).Msgf("agreement not yet in effect: %s", agg.EffectiveAt.String())
+					return fiber.NewError(fiber.StatusBadRequest, FailedToValidateRequest)
+				}
+			}
+
+			if agg.ExpiresAt != nil && !agg.ExpiresAt.IsZero() {
+				if agg.ExpiresAt.Before(time.Now()) {
```
elffjs, line 236: Is this necessarily a problem? If we don't have the access from this particular agreement yet but for the things we're asking for we _do_ have permission. I don't think you want to fail the whole thing.

elffjs:
Also, it seems like in the owner case we should be able to view anything.

elffjs (approved):
The comments that I haven't resolved are things that trouble me, and I think we should address them soon. On the other hand this PR has been open for 10 years and I don't want that to last. This will move things forward.

Allyson-English:
@elffjs @KevinJoiner I hear you both on the concerns-- I started breaking things out into separate pr's yesterday, I just tagged you both in one (to clean up how we make the dex connection and bring the custom claims into this repo).. I'm happy to merge this pr in and then keep iterating in that way!

internal/controllers/token_exchange.go
```diff
@@ -245,26 +211,96 @@ func (t *TokenExchangeController) evaluateSacdDoc(c *fiber.Ctx, record *models.P
...
+				t.logger.Err(err).Str("grantee", grantee.Hex()).Msgf("failed to validate asset did %s in permission agreement", agg.Asset)
+				return fiber.NewError(fiber.StatusBadRequest, FailedToValidateRequest)
+			}
+
+			if err := t.evaluatePermissions(userPermissionGrants, tokenReq); err != nil {
+				t.logger.Err(err).Str("grantee", grantee.Hex()).Msg("failed to evaluate permissions agreement")
+				return fiber.NewError(fiber.StatusBadRequest, FailedToValidateRequest)
+			}
+		}
+	}
```
KevinJoiner, line 264: A bit confused on what this loop is doing. We are creating a list of grants in userGrantMap and then in this loop we are then requiring that every agreement in the the list be "active" (effectiveAt < time.Now() < expiresAt). When we call evaluate we are not using any of the variables in the loop.

Questions 
1. Why do all agreements need to be active for us to pass. I would think you can get a token as long as the active agreements covers the requested info.

2. Why doesn't the userAttGrants or userPermissionGrants check that the granted behavior is effective.

internal/controllers/token_exchange.go
```diff
@@ -245,26 +211,96 @@ func (t *TokenExchangeController) evaluateSacdDoc(c *fiber.Ctx, record *models.P
...
+	userPermissionGrants, userAttGrants, err := t.userGrantMap(record)
+	if err != nil {
+		t.logger.Err(err).Str("grantee", grantee.Hex()).Msg("failed to generate permission grants")
+		return fiber.NewError(fiber.StatusBadRequest)
+	}
+
+	for _, agg := range record.Data.Agreements {
+		switch agg.Type {
+		case "cloudevent":
+			if agg.EventType != cloudevent.TypeAttestation {
```
KevinJoiner, line 223: This check requires all cloudevent agreements to be fore attestations which

KevinJoiner (changes requested):

KevinJoiner:
> @elffjs @KevinJoiner I hear you both on the concerns-- I started breaking things out into separate pr's yesterday, I just tagged you both in one (to clean up how we make the dex connection and bring the custom claims into this repo).. I'm happy to merge this pr in and then keep iterating in that way!

Just saw this message let's do your first approach

PR #34, opened 2025-05-13
ex-3603-token-exchange-break-out-ipfs-svc

Allyson-English:


charts/token-exchange-api/values.yaml
```diff
@@ -34,6 +34,7 @@ env:
   IDENTITY_URL: http://identity-api-dev:8080/query
   IPFS_BASE_URL: https://assets.dimo.xyz/ipfs
   DEV_LICENSE_VALIDATOR_FEATURE_FLAG: true
+  IPFS_TIMEOUT: 30
```
KevinJoiner, line 37: This can be a string that we call [time.ParseDuration](https://pkg.go.dev/time#ParseDuration) on
```suggestion
  IPFS_TIMEOUT: 30s
```

KevinJoiner (changes requested):

internal/services/dex_service.go
```diff
@@ -32,42 +30,33 @@ type PrivilegeTokenDTO struct {
...
 	if err != nil {
-		return nil, conn, err
+		return nil, fmt.Errorf("failed to connect to dex gRPC server: %w", err)
 	}
-	dexClient := dgrpc.NewDexClient(conn)
-	return dexClient, conn, nil
+
+	return &DexClient{
+		log:    log,
+		client: dgrpc.NewDexClient(conn),
```
elffjs, line 41: I'm so happy about this.

This will probably never get closed but it's okay.

internal/services/dex_service.go
```diff
@@ -32,42 +30,33 @@ type PrivilegeTokenDTO struct {
...
-	}
-}
-
-func (d *dexService) getDexGrpcConnection() (dgrpc.DexClient, *grpc.ClientConn, error) {
-	conn, err := grpc.NewClient(d.dexGRPCAddr, grpc.WithTransportCredentials(insecure.NewCredentials()))
+func NewDexClient(log *zerolog.Logger, dexgRPCAddr string) (*DexClient, error) {
+	conn, err := grpc.NewClient(dexgRPCAddr, grpc.WithTransportCredentials(insecure.NewCredentials()))
 	if err != nil {
-		return nil, conn, err
+		return nil, fmt.Errorf("failed to connect to dex gRPC server: %w", err)
```
elffjs, line 36: Contrary to the name, this does not attempt to connect to anything!

charts/token-exchange-api/values.yaml
elffjs: Would be nice if our config loader understood `time.Duration`!

internal/config/settings.go
```diff
@@ -1,19 +1,22 @@
...
+	JWKKeySetURL             string        `yaml:"JWT_KEY_SET_URL"`
+	BlockchainNodeURL        string        `yaml:"BLOCKCHAIN_NODE_URL"`
+	DexGRPCAdddress          string        `yaml:"DEX_GRPC_ADDRESS"`
+	UsersAPIGRPCAddress      string        `yaml:"USERS_API_GRPC_ADDRESS"`
+	ContractAddressWhitelist string        `yaml:"CONTRACT_ADDRESS_WHITELIST"`
+	ContractAddressSacd      string        `yaml:"CONTRACT_ADDRESS_SACD"`
+	IdentityURL              string        `yaml:"IDENTITY_URL"`
+	IPFSBaseURL              string        `yaml:"IPFS_BASE_URL"`
+	DevLiscFeatureFlag       bool          `yaml:"DEV_LICENSE_VALIDATOR_FEATURE_FLAG"`
+	IPFSTimeout              time.Duration `yaml:"IPFS_TIMEOUT"`
```
elffjs, line 21: Could you try this? I'm not sure this is going to work as-is.

Allyson-English: just tried this locally and was able to make requests successfully🫡

elffjs: Hum, did someone add something for...? Maybe our fork?

KevinJoiner: I don't think this works for the env I think you need to use a string.

charts/token-exchange-api/values-prod.yaml
```diff
@@ -16,6 +16,7 @@ env:
   IDENTITY_URL: http://identity-api-prod:8080/query
   IPFS_BASE_URL: https://assets.dimo.xyz/ipfs
   DEV_LICENSE_VALIDATOR_FEATURE_FLAG: false
+  IPFS_TIMEOUT: 30
```
KevinJoiner, line 19:
```suggestion
  IPFS_TIMEOUT: 30s
```

internal/services/custom_claims.go
```diff
@@ -0,0 +1,43 @@
+package services
```
KevinJoiner, line 1: merge in main since this package moved

internal/config/settings.go
elffjs: Try doing

```
export IPFS_TIMEOUT='40s'
go run ./cmd/token-exchange-api
```

I think this will blow up. If you load from a settings file I think it will work because go-yaml had a case for that.

internal/services/ipfs_service.go
```diff
@@ -0,0 +1,83 @@
...
+	"net/url"
+	"strings"
+	"time"
+
+	"github.com/rs/zerolog"
+)
+
+const (
+	DefaultIPFSPrefix  = "ipfs://"
+	DefaultIPFSTimeout = "30s"
```
KevinJoiner, line 17: This doesn't need to be a string
```suggestion
	DefaultIPFSTimeout = time.Second * 30
```

internal/services/ipfs_service.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	resp, err := i.client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("failed to execute request: %w", err)
+	}
+	defer resp.Body.Close() //nolint:errcheck
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("invalid ipfs status code: %d", resp.StatusCode)
```
KevinJoiner, line 79: might be nice to return the body of the response as well if there is a helpful error message telling us why this failed.

internal/config/settings.go
elffjs: https://go.dev/play/p/zI5za3LN3AM

KevinJoiner (changes requested):

internal/config/settings.go
elffjs: We could add some stuff here:

https://github.com/DIMO-Network/shared/blob/1a95c46a8ef25156ed6ef42b12f2a0dd4e4ebcdc/pkg/settings/config_loader.go#L135

internal/controllers/token_exchange.go
```diff
@@ -24,6 +20,11 @@ import (
 	"github.com/rs/zerolog"
 )
 
+//go:generate mockgen -source ./token_exchange.go -destination mocks/token_exchange_mock.go -package mock_controller_test
```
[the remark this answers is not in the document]
elffjs: Interesting point. I do like it.

internal/services/ipfs_service.go
```diff
@@ -0,0 +1,79 @@
...
+const DefaultIPFSPrefix = "ipfs://"
+
+type IPFSClient struct {
+	logger      *zerolog.Logger
+	client      *http.Client
+	ipfsBaseURL *url.URL
+	timeout     time.Duration
+}
+
+func NewIPFSClient(logger *zerolog.Logger, ipfsBaseURL string, ipfsTimeout int) (*IPFSClient, error) {
```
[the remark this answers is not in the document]
elffjs: I like not just passing in a `Settings` struct to every level of the app! Much tidier.

internal/controllers/token_exchange_test.go
```diff
@@ -87,7 +90,7 @@ func TestTokenExchangeController_GetDeviceCommandPermissionWithScope(t *testing.
 			},
 			mockSetup: func() {
 				contractsMgr.EXPECT().GetSacd(c.settings.ContractAddressSacd, &client).Return(mockSacd, nil)
-
+				mockipfs.EXPECT().Fetch(gomock.Any(), gomock.Any()).Return(nil, errors.New("no valid doc"))
```
elffjs, line 93: nitpick: I'm surprised we would need to do `Any` here.

internal/services/custom_claims.go
```diff
@@ -0,0 +1,43 @@
...
+func (c *CustomClaims) Proto() (*structpb.Struct, error) {
+	ap := make([]any, len(c.PrivilegeIDs))
+
+	for i := range c.PrivilegeIDs {
+		ap[i] = int64(c.PrivilegeIDs[i])
+	}
+
+	return structpb.NewStruct(
+		map[string]any{
+			"contract_address": hexutil.Encode(c.ContractAddress[:]),
```
elffjs, line 33: note: We should probably change this to `c.ContractAddress.Hex()` at some point.

That is a breaking change, if any clients were doing the (naughty) thing of just using a string comparison.

elffjs (changes requested):
Let's fix the config loading somehow. Otherwise, this is great. Very tidy.

internal/services/custom_claims.go
elffjs: @Allyson-English I think we still need to do this?

internal/services/ipfs_service.go
```diff
@@ -24,24 +24,26 @@ type IPFSClient struct {
...
 	if err != nil {
 		return nil, fmt.Errorf("invalid IPFS base URL: %w", err)
 	}
 
-	if ipfsTimeout.Seconds() == 0 {
-		ipfsTimeout, err = time.ParseDuration(DefaultIPFSTimeout)
-		if err != nil {
-			return nil, err
-		}
+	timout, err := time.ParseDuration(ipfsTimeout)
```
elffjs, line 33: 👍

internal/services/custom_claims.go
elffjs: I think on main this has been moved to [`pkg/tokenclaims`](https://github.com/DIMO-Network/token-exchange-api/tree/main/pkg/tokenclaims).

elffjs (approved):

PR #36, opened 2025-05-14
ex-3584-add-attestations-to-custom-claims

Allyson-English:


pkg/tokenclaims/tokenclaims.go
```diff
@@ -16,6 +16,16 @@ type CustomClaims struct {
 	ContractAddress common.Address         `json:"contract_address"`
 	TokenID         string                 `json:"token_id"`
 	PrivilegeIDs    []privileges.Privilege `json:"privilege_ids"`
+	CloudEvents     *CloudEvent            `json:"cloud_event"`
+}
+
+type CloudEvent struct {
```
elffjs, line 22: I feel like we need `json` tags here and in `Attestation`.

pkg/tokenclaims/tokenclaims.go
```diff
@@ -16,6 +16,16 @@ type CustomClaims struct {
 	ContractAddress common.Address         `json:"contract_address"`
 	TokenID         string                 `json:"token_id"`
 	PrivilegeIDs    []privileges.Privilege `json:"privilege_ids"`
+	CloudEvents     *CloudEvent            `json:"cloud_event"`
```
elffjs, line 19: It’s jarring to see both plural and singular here.

pkg/tokenclaims/tokenclaims.go
```diff
@@ -37,6 +47,7 @@ func (c *CustomClaims) Proto() (*structpb.Struct, error) {
 			"contract_address": hexutil.Encode(c.ContractAddress[:]),
 			"token_id":         c.TokenID,
 			"privilege_ids":    ap,
+			"cloud_events":     c.CloudEvents,
```
elffjs, line 50: I’m worried that this won’t serialize. You may have to do another protobuf struct.

elffjs (changes requested):
It looks like we’re just passing through for now, That’s fine. I am worried about the pluralization, and the serialization for gRPC.

KevinJoiner (changes requested):

pkg/tokenclaims/tokenclaims.go
```diff
@@ -16,16 +16,17 @@ type CustomClaims struct {
 	ContractAddress common.Address         `json:"contract_address"`
 	TokenID         string                 `json:"token_id"`
 	PrivilegeIDs    []privileges.Privilege `json:"privilege_ids"`
-	CloudEvents     *CloudEvent            `json:"cloud_event"`
+	CloudEvents     *CloudEvents           `json:"cloud_event"`
```
elffjs, line 19: Sorry: Plural or singular? This doesn't seem to match line 66, and see my next comment.

pkg/tokenclaims/tokenclaims.go
```diff
@@ -16,16 +16,17 @@ type CustomClaims struct {
...
-	Attestations []Attestation
+type CloudEvents struct {
+	Events []Event `json:"events"`
 }
 
-type Attestation struct {
-	Source *string
-	IDs    []string
+type Event struct {
+	EventType string   `json:"eventType"`
```
elffjs, line 27: I worry we're trying to make this structure do too much. We're trying to use it as

1. Part of the body of the HTTP request to this API. There the convention is camel case, which we seem to have here.
2. Some internal data transfer object.
3. As a struct that importers of `pkg/tokenclaims` can use to deserialize JWTs. To be consistent there, the convention is underscores, and indeed on line 66 we have `cloud_events`.

pkg/tokenclaims/tokenclaims.go
```diff
@@ -42,12 +43,27 @@ func (c *CustomClaims) Proto() (*structpb.Struct, error) {
 		ap[i] = int64(c.PrivilegeIDs[i])
 	}
 
+	ces := make((map[string]map[string]any))
+	for _, evt := range c.CloudEvents.Events {
+		if _, ok := ces[evt.EventType]; !ok {
+			ces[evt.EventType] = map[string]any{}
+		}
+
+		if _, ok := ces[evt.EventType][*evt.Source]; !ok {
```
elffjs, line 52: Is this dereference safe?

pkg/tokenclaims/tokenclaims.go
```diff
@@ -42,12 +43,27 @@ func (c *CustomClaims) Proto() (*structpb.Struct, error) {
...
+			// NOTE: this will overwrite, to avoid that, I think that prior
+			// to this point we should return an error if someone is passing
+			// multiple cloud event requests with the same source, directing them that
```
elffjs: That's a big decision! I think it's fine to make the choice but it does constrain us. Curious what others think.

elffjs (changes requested):

pkg/tokenclaims/tokenclaims.go
Allyson-English: this comment is from a previous version! towards the end of the day yesterday I revised what was happening and took that out

pkg/tokenclaims/tokenclaims.go
Allyson-English: yeah.. i hear that point. I wanted to avoid making two structs that we almost identical except for json tags because I thought that would be confusing/ I wanted to make it clear what we're expecting to be passed in.. I'm open to other ideas on how to best accomplish that. I guess separating the event struct in the token claims and what we're using the request is a good place to start?

pkg/tokenclaims/tokenclaims.go
```diff
@@ -32,11 +45,33 @@ func (c *CustomClaims) Proto() (*structpb.Struct, error) {
 		ap[i] = int64(c.PrivilegeIDs[i])
 	}
 
+	ces := make((map[string]map[string][]any))
+	for _, evt := range c.CloudEvents.Events {
```
elffjs, line 49: Can't `c.CloudEvents` be nil? This looks dangerous.

pkg/tokenclaims/tokenclaims.go
elffjs: I think mixing casing is going to be rough. At this point we control both sides of this thing so it's okay to make a choice without full conviction.

We're going to be redoing these a bit anyway, so now is the time to say "I think these should be camel" and we can just include both styles for the existing fields and eventually get rid.

pkg/tokenclaims/tokenclaims.go
```diff
@@ -32,11 +45,33 @@ func (c *CustomClaims) Proto() (*structpb.Struct, error) {
...
+		if _, ok := ces[evt.EventType]; !ok {
+			ces[evt.EventType] = map[string][]any{}
+		}
+
+		source := evt.Source
+		if source == nil {
+			source = &GlobalAttestationPermission
```
elffjs, line 56: I'm worried about someone passing this in, like making a SACD doc with `"source": "GLOBAL_ATTESTATION_PERMISSION"` and that becomes part of your spec.

Not blocking, just worried.

pkg/tokenclaims/tokenclaims.go
elffjs: This is my only blocking comment.

elffjs (changes requested):

pkg/tokenclaims/tokenclaims.go
Allyson-English: I'm less worried about that happening accidentally because when we're ingesting in dis we check for a valid cloud event header which includes `common.IsHexAddress(eventHdr.Source)` 

I do see how this makes things a little bit confusing though... my main goal was to be able to easily distinguish what type of permission we were checking for and I wanted to be as explicit as possible

elffjs (approved):

pkg/tokenclaims/tokenclaims.go
```diff
@@ -32,11 +45,34 @@ func (c *CustomClaims) Proto() (*structpb.Struct, error) {
...
+				source = &GlobalAttestationPermission
+			}
+
+			if _, ok := ces[evt.EventType][*source]; !ok {
+				ces[evt.EventType][*source] = []any{}
+			}
+
+			for _, id := range evt.IDs {
+				ces[evt.EventType][*source] = append(ces[evt.EventType][*source], id)
+			}
```
KevinJoiner, line 66: Can we flip this? I don't love having `GlobalAttestationPermission` as a var because that means any package that is imported in token exchange has the ability to change the value. secondarily it's nice to not have to dereference the pointer every time we use it.
```suggestion
            const GlobalAttestationPermission = "GLOBAL_ATTESTATION_PERMISSION"
            // .
            // .
            // .
			source := GlobalAttestationPermission
			if source != nil {
				source = *evt.Source
			}

			if _, ok := ces[evt.EventType][source]; !ok {
				ces[evt.EventType][source] = []any{}
			}

			for _, id := range evt.IDs {
				ces[evt.EventType][source] = append(ces[evt.EventType][source], id)
			}
```

internal/controllers/token_exchange.go
```diff
@@ -58,6 +59,8 @@ type PermissionTokenRequest struct {
 	NFTContractAddress string `json:"nftContractAddress" example:"0xbA5738a18d83D41847dfFbDC6101d37C69c9B0cF" validate:"required"`
 	// Audience is the intended audience for the token.
 	Audience []string `json:"audience" validate:"optional"`
+	// CloudEvent request, includes attestations
+	CloudEvents *tokenclaims.CloudEvents `json:"cloudEvents"`
```
KevinJoiner, line 63: I would be down to use a different type here so that we are not mixing camelCase and snake_case. Even though it is easier for us to use the same struct. I think it's more important that the end user has a consistent experience without having to switch between snake and camel case.

KevinJoiner (changes requested):

pkg/tokenclaims/tokenclaims.go
```diff
@@ -32,11 +45,34 @@ func (c *CustomClaims) Proto() (*structpb.Struct, error) {
 		ap[i] = int64(c.PrivilegeIDs[i])
 	}
 
+	ces := make((map[string]map[string][]any))
```
elffjs, line 48: I think an interesting "test" here would be to just have the protobuf serializer try to serialize the output here and see if it fails.

elffjs (changes requested):
I'm happy with it except for the two things marked with ☡. Let's fix (or dismiss!) those and merge away.

internal/controllers/token_exchange.go
```diff
@@ -84,27 +97,27 @@ func NewTokenExchangeController(logger *zerolog.Logger, settings *config.Setting
 // @Summary     The authenticated user must have a confirmed Ethereum address with those
 // @Summary     privileges on the correct token.
 // @Accept      json
-// @Param       tokenRequest body controllers.PermissionTokenRequest true "Requested privileges: must include address, token id, and privilege ids"
+// @Param       tokenRequest body controllers.TokenRequest true "Requested privileges: must include address, token id, and privilege ids"
 // @Produce     json
-// @Success     200 {object} controllers.PermissionTokenResponse
+// @Success     200 {object} controllers.TokenRequest
```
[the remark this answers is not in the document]
Allyson-English: I changed this because we have a specific meaning for permissions and this isn't just controlling that anymore... happy to change back

pkg/tokenclaims/tokenclaims.go
```diff
@@ -32,11 +45,34 @@ func (c *CustomClaims) Proto() (*structpb.Struct, error) {
...
+	if c.CloudEvents != nil {
+		for _, evt := range c.CloudEvents.Events {
+			if _, ok := ces[evt.EventType]; !ok {
+				ces[evt.EventType] = map[string][]any{}
+			}
+
+			source := GlobalAttestationPermission
```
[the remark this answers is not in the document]
KevinJoiner: Do we think it would be better to use empty string here? I am torn because I get your point of not wanting to add a special string to token because with that logic we should just go back to using `*`. On the other hand empty string seems scary because I could see it being handled incorrectly by someone else.
```suggestion
			source := ""
```

pkg/tokenclaims/tokenclaims.go
```diff
@@ -11,11 +11,24 @@ import (
 	"google.golang.org/protobuf/types/known/structpb"
 )
 
+var GlobalAttestationPermission = "GLOBAL_ATTESTATION_PERMISSION"
```
KevinJoiner, line 14:
```suggestion
const GlobalAttestationPermission = "GLOBAL_ATTESTATION_PERMISSION"
```

KevinJoiner (changes requested):

internal/controllers/token_exchange.go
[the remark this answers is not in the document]
elffjs: We resolved this conversation, but I think what's here is still wrong.

I will fix it, but I think one should just respond to comments so everyone understands something's been done, even the PR author.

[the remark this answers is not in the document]
elffjs: As with the request, I took out the name permission so it would be more neutral.

internal/controllers/token_exchange.go
```diff
@@ -128,26 +141,40 @@ func (t *TokenExchangeController) GetDeviceCommandPermissionWithScope(c *fiber.C
 		t.logger.Warn().Err(err).Msg("Failed to get valid SACD document")
 		// If the user doesn't have a valid IPFS doc, check bitstring
 		// We call the contract again because this handles the case where the caller is the owner of the asset.
-		return t.evaluatePermissionsBits(c, s, nftAddr, pr, ethAddr)
+		return t.evaluatePermissionsBits(c, s, nftAddr, tokenReq, ethAddr)
```
elffjs, line 144: I just noticed we're passing around the pointer for no great reason here. Not a problem for this PR, not created in this PR, but I do want to fix this.

internal/controllers/token_exchange.go
```diff
@@ -60,9 +61,21 @@ type PermissionTokenRequest struct {
 	NFTContractAddress string `json:"nftContractAddress" example:"0xbA5738a18d83D41847dfFbDC6101d37C69c9B0cF" validate:"required"`
 	// Audience is the intended audience for the token.
 	Audience []string `json:"audience" validate:"optional"`
+	// CloudEvent request, includes attestations
+	CloudEvents *CloudEvents `json:"cloudEvents"`
```
elffjs, line 65: Idle comment: I think this pointer is probably not doing any good.

internal/controllers/token_exchange.go
```diff
@@ -84,27 +97,27 @@ func NewTokenExchangeController(logger *zerolog.Logger, settings *config.Setting
...
 
-	nftAddr := common.HexToAddress(pr.NFTContractAddress)
+	nftAddr := common.HexToAddress(tokenReq.NFTContractAddress)
 
-	t.logger.Debug().Interface("request", pr).Msg("Got request.")
+	t.logger.Debug().Interface("request", tokenReq).Msg("Got request.")
 
-	if len(pr.Privileges) == 0 {
-		return fiber.NewError(fiber.StatusBadRequest, "Please provide at least one privilege.")
+	if len(tokenReq.Privileges) == 0 && tokenReq.CloudEvents == nil {
```
elffjs: ☡ The list contained in `CloudEvents` could still be empty, and I think we would be similarly upset then; so we should probably do a length check here too.

elffjs: I'm just gonna add it so I can get to yes. Forgive me.

elffjs (approved):

KevinJoiner (approved):

PR #37, opened 2025-05-14
ex-3544-sacd-extended-attestation-permissions

Allyson-English:


internal/controllers/token_exchange.go
```diff
@@ -210,30 +214,109 @@ func (t *TokenExchangeController) evaluateSacdDoc(c *fiber.Ctx, record *models.P
...
+
+			if err := t.evaluatePermissions(userPermGrants, tokenReq); err != nil {
+				logger.Err(err).Msg("failed to evaluate permissions agreement")
+				return fiber.NewError(fiber.StatusBadRequest, "failed to validate request")
+			}
+		}
+	}
+
+	// If we get here, all permission and attestation claims are valid
+	return t.createAndReturnToken(c, tokenReq, grantee)
```
KevinJoiner, line 261: We already loop through the agreements in userGrantMap we should not need to do this again.
We can move the validation of assets into userGrantMap since there is already a type switch in there. and then we can have the following for this function
```suggestion
	userPermGrants, cloudEvtGrants := t.userGrantMap(record
                
    if err := t.evaluatePermissions(userPermGrants, tokenReq); err != nil {
		logger.Err(err).Msg("failed to evaluate permissions agreement")
		return fiber.NewError(fiber.StatusBadRequest, "failed to validate request")
	}		
	if err := t.evaluateCloudEvents(cloudEvtGrants, tokenReq); err != nil {
		logger.Err(err).Msg("failed to validate request")
		return fiber.NewError(fiber.StatusBadRequest, "failed to validate request")
	}
	return t.createAndReturnToken(c, tokenReq, grantee)
}

	// If we get here, all permission and attestation claims are valid
	return t.createAndReturnToken(c, tokenReq, grantee)
```

internal/controllers/token_exchange.go
```diff
@@ -250,59 +333,50 @@ func (t *TokenExchangeController) evaluateSacdDoc(c *fiber.Ctx, record *models.P
 
 	// If any permissions are missing, return an error
 	if len(missingPermissions) > 0 {
-		return fiber.NewError(fiber.StatusBadRequest,
-			fmt.Sprintf("Address %s lacks permissions %v on token id %d for asset %s.",
-				grantee.Hex(), missingPermissions, pr.TokenID, pr.NFTContractAddress))
+		return fmt.Errorf("missing permissions: %v on token id %d for asset %s", missingPermissions, tokenReq.TokenID, tokenReq.NFTContractAddress)
```
KevinJoiner, line 336: having the grantee in the error message isn't bad

internal/controllers/token_exchange.go
```diff
@@ -210,30 +214,109 @@ func (t *TokenExchangeController) evaluateSacdDoc(c *fiber.Ctx, record *models.P
...
+			// Validate the asset DID if it exists in the record
+			valid, err := t.validateAssetDID(agg.Asset, tokenReq)
+			if err != nil || !valid {
+				logger.Err(err).Msgf("failed to validate asset did %s in permission agreement", agg.Asset)
+				return fiber.NewError(fiber.StatusBadRequest, "failed to validate request")
+			}
+
+			if err := t.evaluatePermissions(userPermGrants, tokenReq); err != nil {
+				logger.Err(err).Msg("failed to evaluate permissions agreement")
+				return fiber.NewError(fiber.StatusBadRequest, "failed to validate request")
```
KevinJoiner, line 255: The error message contains what permissions the user is missing which I think we should continue to bubble up sine it is helpful for devs.

internal/controllers/token_exchange.go
```diff
@@ -250,59 +333,50 @@ func (t *TokenExchangeController) evaluateSacdDoc(c *fiber.Ctx, record *models.P
...
+	return nil
 }
 
-func intArrayTo2BitArray(indices []int64, length int) (*big.Int, error) {
-	mask := big.NewInt(0)
-
-	for _, index := range indices {
-		if index < 0 || index >= int64(length) {
-			return big.NewInt(0), fmt.Errorf("invalid index %d. These must be non-negative and less than %d", index, length)
+func (t *TokenExchangeController) evaluateCloudEvents(agreement map[string]map[string]*shared.StringSet, tokenReq *PermissionTokenRequest) error {
```
KevinJoiner, line 342:
```suggestion
func evaluateCloudEvents(agreement map[string]map[string]*shared.StringSet, tokenReq *PermissionTokenRequest) error {
```

internal/controllers/token_exchange.go
```diff
@@ -250,59 +333,50 @@ func (t *TokenExchangeController) evaluateSacdDoc(c *fiber.Ctx, record *models.P
...
+		if _, ok := grantedAggs[*source]; !ok {
+			err = errors.Join(err, fmt.Errorf("lacking %s grant for requested source: %s", req.EventType, *source))
+			continue
+		}
 
-	requestNFTAddr := common.HexToAddress(req.NFTContractAddress)
+		// NOTE: do we want to explicitly enforce that
+		// someone has to ask for the exact ids they've been granted?
+		// this is assuming we do not
+		if len(grantedAggs[*source].Slice()) == 0 {
```
KevinJoiner, line 364:
```suggestion
		if grantedAggs[*source].Len() == 0 {
```

KevinJoiner (changes requested):

internal/controllers/token_exchange.go
Allyson-English: I kept this in because of the odd situation with validating the asset:
- For the SACD permission grant the asset is inside the permission data but 
- for attestations it's inside the agreements (which is in array).... 

I'm wondering if it makes the most sense the remove the possibility of having a separate asset listed in each attestation agreement (I believe we initially did this for maximum flexibility... but now that I'm working on this change I don't know that there's a concrete use case for it)

elffjs: I think as presently constructed there's no way the assets can differ. I would just not handle this case—the top level should have the asset, that's fine.

elffjs: We seem to be really confusing this amongst ourselves. I think @LorranSutter kinda waffles on this too: in the example [permissions doc](https://github.com/DIMO-network/sacd) there's a top-level asset but none in the agreement, but [the README](https://github.com/DIMO-network/sacd) does not allow for this possibility.

elffjs: The way these are currently put on-chain, there's really no way for the asset to vary, and assets basically have to be attached to an NFT. I would honestly just assume and verify a top-level `asset` directive and forget about the rest for now.

Keep in mind that these two types of agreements might be in the same document.

internal/controllers/token_exchange.go
```diff
@@ -276,59 +278,51 @@ func (t *TokenExchangeController) evaluateSacdDoc(c *fiber.Ctx, record *models.P
...
-//
-// Returns:
-//   - bool: true if the DID is valid and matches the request parameters, false otherwise
-//   - error: An error describing why validation failed, or nil if validation succeeded
-func (t *TokenExchangeController) validateAssetDID(did string, req *TokenRequest) (bool, error) {
-	decodedDID, err := cloudevent.DecodeNFTDID(did)
-	if err != nil {
-		return false, fmt.Errorf("failed to decode DID: %w", err)
-	}
+		idSet, ok := grantedAggs[*source]
```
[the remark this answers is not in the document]
elffjs: I think there are perhaps more pointers flying around than need exist.

internal/controllers/token_exchange.go
```diff
@@ -276,59 +278,51 @@ func (t *TokenExchangeController) evaluateSacdDoc(c *fiber.Ctx, record *models.P
...
+		if !ok {
+			err = errors.Join(err, fmt.Errorf("lacking %s grant for requested source: %s", req.EventType, *source))
+			continue
+		}
 
-	requestNFTAddr := common.HexToAddress(req.NFTContractAddress)
+		// NOTE: do we want to explicitly enforce that
```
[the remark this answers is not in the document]
elffjs: If we had a separate permission for _listing_, that would be interesting.

elffjs (changes requested):
Only the ☡ are super serious. They're corner cases but this seems like the right PR in which to stamp these out.

I will take a closer look at the tests. I like the tests.

internal/controllers/sacd_helpers.go
```diff
@@ -0,0 +1,129 @@
...
+			if agreement.Source == nil {
+				source = &tokenclaims.GlobalAttestationPermission
+			}
+
+			if _, ok := cloudEvtGrants[agreement.EventType][*source]; !ok {
+				cloudEvtGrants[agreement.EventType][*source] = shared.NewStringSet()
+			}
+
+			for _, id := range agreement.ID {
+				cloudEvtGrants[agreement.EventType][*source].Add(id)
```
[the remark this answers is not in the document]
Allyson-English: What do you see as the correct behavior? Both of those can't be true so it seems like our only options would be:
- evaluate sequentially, as done here
- always opt for the most restrictive option 
I can see arguments for both.

internal/controllers/token_exchange.go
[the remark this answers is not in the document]
Allyson-English: if we aren't going to let people specify ids along with the super-source, we'd probably want to go back and revise what the allowable structure of a grant looks like

internal/controllers/sacd_helpers.go
```diff
@@ -0,0 +1,127 @@
...
+
+			if _, ok := cloudEvtGrants[agreement.EventType][source]; !ok {
+				cloudEvtGrants[agreement.EventType][source] = shared.NewStringSet()
+			}
+
+			for _, id := range agreement.IDs {
+				cloudEvtGrants[agreement.EventType][source].Add(id)
+			}
+
+		case "permissions":
```
LorranSutter, line 69: Remember it is "permission" instead of "permissions"

LorranSutter (changes requested):

internal/controllers/sacd_helpers.go
```diff
@@ -0,0 +1,127 @@
...
+				if time.Now().Before(*agreement.EffectiveAt) {
+					return nil, nil, errors.New("agreement not yet in effect")
+				}
+			}
+
+			if agreement.ExpiresAt != nil && !agreement.ExpiresAt.IsZero() {
+				if agreement.ExpiresAt.Before(time.Now()) {
+					return nil, nil, errors.New("agreement expired")
+				}
+			}
```
KevinJoiner: 1. I don't think we need to check the expiredAt and effectiveAt fields twice.
2 We don't need to error if the agreement is expired just don't add it to the map. The SACD should be able to have more than one cloudevent agreement and it's okay if one of them is expired.

internal/controllers/sacd_helpers.go
```diff
@@ -0,0 +1,127 @@
...
+		if agreement.ExpiresAt != nil {
+			if agreement.ExpiresAt.Before(time.Now()) {
+				continue
+			}
+		}
+
+		switch agreement.Type {
+		case "cloudevent":
+			if valid, err := validateAssetDID(agreement.Asset, nftAddr, tokenID); err != nil || !valid {
+				return nil, nil, fmt.Errorf("failed to validate attestation asset: %s", agreement.Asset)
```
KevinJoiner, line 34: If the asset has to be the same as the tokenId is there any benefit to having it in the agreement? Also Why do we care about it for cloudevent only but not for permissions? 
If it is just for some future use case then I think we should not have it in the agreement and just check `record.Data.Asset` in all cases.

internal/controllers/token_exchange.go
```diff
@@ -276,59 +278,48 @@ func (t *TokenExchangeController) evaluateSacdDoc(c *fiber.Ctx, record *models.P
...
-		return false, fmt.Errorf("failed to decode DID: %w", err)
-	}
+		idSet, ok := grantedAggs[source]
+		if !ok {
+			err = errors.Join(err, fmt.Errorf("lacking %s grant for requested source: %s", req.EventType, source))
+			continue
+		}
 
-	requestNFTAddr := common.HexToAddress(req.NFTContractAddress)
+		if idSet.Len() == 0 {
```
KevinJoiner: We should treat this the same as source where `*` means all so we are consistent.
```suggestion
		if idSet.Contains(tokenclaims.GlobalAttestationPermission) {
```

internal/controllers/token_exchange.go
```diff
@@ -276,59 +278,48 @@ func (t *TokenExchangeController) evaluateSacdDoc(c *fiber.Ctx, record *models.P
...
+			continue
+		}
 
-	if decodedDID.ContractAddress != requestNFTAddr {
-		return false, fmt.Errorf("DID contract address %s does not match request contract address %s",
-			decodedDID.ContractAddress.Hex(), requestNFTAddr.Hex())
-	}
+		if len(req.IDs) == 0 {
+			err = errors.Join(err, fmt.Errorf("requesting global access to %s cloudevents for %s but only granted subset", source, req.EventType))
+		}
```
KevinJoiner: If we use star then this error would just say that the user didn't request any events.

pkg/tokenclaims/tokenclaims.go
```diff
@@ -11,7 +11,7 @@ import (
 	"google.golang.org/protobuf/types/known/structpb"
 )
 
-var GlobalAttestationPermission = "GLOBAL_ATTESTATION_PERMISSION"
+const GlobalAttestationPermission = "GLOBAL_ATTESTATION_PERMISSION"
```
KevinJoiner:
```suggestion
const GlobalAttestationPermission = "*"
```

internal/controllers/token_exchange.go
```diff
@@ -276,59 +279,58 @@ func (t *TokenExchangeController) evaluateSacdDoc(c *fiber.Ctx, record *models.P
...
-	return t.createAndReturnToken(c, pr, grantee)
+	return nil
 }
 
-func intArrayTo2BitArray(indices []int64, length int) (*big.Int, error) {
-	mask := big.NewInt(0)
+func evaluateCloudEvents(agreement map[string]map[string]*shared.StringSet, tokenReq *TokenRequest) error {
+	var err error
+	for _, req := range tokenReq.CloudEvents.Events {
+		if !common.IsHexAddress(req.Source) && req.Source != "*" {
```
KevinJoiner, line 291:
```suggestion
		if !common.IsHexAddress(req.Source) && req.Source != tokenclaims.CloudEventTypeGlobal {
```

internal/controllers/token_exchange.go
```diff
@@ -276,59 +279,58 @@ func (t *TokenExchangeController) evaluateSacdDoc(c *fiber.Ctx, record *models.P
...
-//
-// Returns:
-//   - bool: true if the DID is valid and matches the request parameters, false otherwise
-//   - error: An error describing why validation failed, or nil if validation succeeded
-func (t *TokenExchangeController) validateAssetDID(did string, req *TokenRequest) (bool, error) {
-	decodedDID, err := cloudevent.DecodeNFTDID(did)
-	if err != nil {
-		return false, fmt.Errorf("failed to decode DID: %w", err)
-	}
+		if globalAggs, grantedAll := grantedAggs[tokenclaims.CloudEventTypeGlobal]; grantedAll {
```
KevinJoiner, line 306: nit: I always like to use ok for something existing in a map. But this is highly debatable so you can ignore
```suggestion
		if globalAggs, ok := grantedAggs[tokenclaims.CloudEventTypeGlobal]; ok {
```

internal/controllers/token_exchange.go
```diff
@@ -276,59 +279,58 @@ func (t *TokenExchangeController) evaluateSacdDoc(c *fiber.Ctx, record *models.P
...
+			if globalAggs.Contains(tokenclaims.CloudEventTypeGlobal) {
+				continue
```
KevinJoiner, line 308:
```suggestion
				continue
```

internal/controllers/sacd_helpers.go
```diff
@@ -0,0 +1,115 @@
...
+		case "cloudevent":
+			if err := validAssetDID(agreement.Asset, nftAddr, tokenID); err != nil {
+				return nil, nil, fmt.Errorf("failed to validate attestation asset: %s", agreement.Asset)
+			}
+
+			if agreement.EffectiveAt != nil && !agreement.EffectiveAt.IsZero() && now.Before(*agreement.EffectiveAt) {
+				continue // agreement not yet in effective
+			}
+
+			if agreement.ExpiresAt != nil && !agreement.ExpiresAt.IsZero() && agreement.ExpiresAt.Before(now) {
```
KevinJoiner, line 41: why are we checking this again here?

internal/controllers/sacd_helpers.go
```diff
@@ -0,0 +1,115 @@
...
+				return nil, nil, errors.New("agreement expired")
+			}
+
+			if _, ok := cloudEvtGrants[agreement.EventType]; !ok {
+				cloudEvtGrants[agreement.EventType] = map[string]*shared.StringSet{}
+			}
+
+			source := strings.ToLower(agreement.Source)
```
KevinJoiner, line 49: I don't think we want a case insensitive check. We want to be using a real address
```suggestion
			source := strings.ToLower(agreement.Source)
```

internal/controllers/token_exchange.go
```diff
@@ -276,59 +279,58 @@ func (t *TokenExchangeController) evaluateSacdDoc(c *fiber.Ctx, record *models.P
...
+		idSet, ok := grantedAggs[source]
+		if !ok {
+			err = errors.Join(err, fmt.Errorf("lacking %s grant for requested source: %s", req.EventType, source))
+			continue
+		}
 
-	if int64(decodedDID.TokenID) != req.TokenID {
-		return false, fmt.Errorf("DID token ID %d does not match request token ID %d",
-			decodedDID.TokenID, req.TokenID)
+		for _, reqID := range req.IDs {
```
KevinJoiner, line 326: if they have and id set of `["id1", "id2", "*"]` and request `["id3"]` we should let them through
```go
if idSet.Contains(tokenclaims.CloudEventTypeGlobal) {
  continue
}
```

KevinJoiner (changes requested):

internal/controllers/sacd_helpers.go
Allyson-English: I agree with you/ raised this earlier to @elffjs I think it's for a future use case... if we're comfortable with not checking it rn I'm good with that. I'd also be happy to remove if from the sacd attestation spec for now (to be added back later once there's a clear use case)

KevinJoiner: Sorry I forgot about the previous conversation on this. Let's remove it from the agreement sections for attestations since it's easier to add things then it is to remove them later.

elffjs: I think the token id should be in the thing that they sign, so it should be in the document. But at this point having it in the individual agreements seems odd (unless those are the only things that the parties sign).

internal/controllers/sacd_helpers.go
```diff
@@ -0,0 +1,146 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to decode DID: %w", err)
+	}
+
+	if decodedDID.ContractAddress != common.HexToAddress(nftContractAddr) {
+		return fmt.Errorf("DID contract address %s does not match request contract address %s",
+			decodedDID.ContractAddress.Hex(), nftContractAddr)
+	}
+
+	if int64(decodedDID.TokenID.Int64()) != tokenID {
```
elffjs, line 126: I'm going to flip this around to
```suggestion
	if decodedDID.TokenID.Cmp(big.NewInt(tokenID)) != 0 {
```
if that's okay.

elffjs: Side note: I don't think we need to cast here, actually.

internal/controllers/sacd_helpers.go
```diff
@@ -0,0 +1,148 @@
...
+	if len(req.IDs) == 0 {
+		return fmt.Errorf("must request at least one cloudevent id or global access request (%s)", tokenclaims.GlobalIdentifier)
+	}
+
+	grantedAggs, ok := agreement[req.EventType]
+	if !ok {
+		return fmt.Errorf("lacking grant for requested event type: %s", req.EventType)
+	}
+
+	globalGrantIDs, ok := grantedAggs[tokenclaims.GlobalIdentifier]
```
elffjs, line 32: Here's what freaks me out about the section that starts here: we use both `ok` and `globalGrantIDs == nil` for basically the same purpose. It almost looks like a typo. I don't have a great answer because this set type is hard to work with.

internal/controllers/sacd_helpers.go
```diff
@@ -0,0 +1,148 @@
...
+	if ok && globalGrantIDs.Contains(tokenclaims.GlobalIdentifier) {
+		return nil
+	}
+
+	sourceGrantIDs, ok := grantedAggs[req.Source]
+	if globalGrantIDs == nil && !ok {
+		return fmt.Errorf("no %s grants for source: %s", req.EventType, req.Source)
+	}
```
elffjs, line 40: Are these lines for optimization and better error messages only, or are they load-bearing?

I think I know the answer but I'm curious what the author thinks.

internal/controllers/sacd_helpers.go
```diff
@@ -0,0 +1,148 @@
...
+
+	if missingIDs := evaluateIDsByGrantSource(globalGrantIDs, sourceGrantIDs, req.IDs); len(missingIDs) > 0 {
+		return fmt.Errorf("lacking %s grant for source %s with ids: %s", req.EventType, req.Source, strings.Join(missingIDs, ","))
+	}
+
+	return nil
+}
+
+func evaluateIDsByGrantSource(globalGrants *set.StringSet, sourceGrants *set.StringSet, requestedIDs []string) []string {
```
elffjs, line 49: You're spending most of your time here handling nils with care. What do you think of this?

https://github.com/DIMO-Network/token-exchange-api/pull/50

elffjs (approved):

internal/controllers/token_exchange.go
```diff
@@ -228,6 +229,7 @@ func (t *TokenExchangeController) getValidSacdDoc(ctx context.Context, source st
 //     nil if all permissions are valid and the token is successfully created and returned
 func (t *TokenExchangeController) evaluateSacdDoc(c *fiber.Ctx, record *models.PermissionRecord, tokenReq *TokenRequest, grantee *common.Address) error {
```
elffjs, line 230: I have a branch here that gets rid of all these `*common.Address` types:

https://github.com/DIMO-Network/token-exchange-api/pull/47

internal/controllers/token_exchange.go
```diff
@@ -138,6 +136,9 @@ func (t *TokenExchangeController) GetDeviceCommandPermissionWithScope(c *fiber.C
 
 	record, err := t.getValidSacdDoc(c.Context(), resPermRecord.Source)
 	if err != nil {
+		if tokenReq.CloudEvents != nil {
```
elffjs, line 139: I have a branch here that gets rid of the `nil` and adds Swagger docs for the fields

https://github.com/DIMO-Network/token-exchange-api/pull/48/files

Feel free to use some version of this

LorranSutter (approved):

internal/controllers/token_exchange.go
```diff
@@ -276,59 +278,22 @@ func (t *TokenExchangeController) evaluateSacdDoc(c *fiber.Ctx, record *models.P
...
-	mask := big.NewInt(0)
-
-	for _, index := range indices {
-		if index < 0 || index >= int64(length) {
-			return big.NewInt(0), fmt.Errorf("invalid index %d. These must be non-negative and less than %d", index, length)
+func evaluateCloudEvents(agreement map[string]map[string]*set.StringSet, tokenReq *TokenRequest) error {
+	var err error
+	for _, req := range tokenReq.CloudEvents.Events {
+		if ceErr := evaluateCloudEvent(agreement, req); ceErr != nil {
+			err = errors.Join(err, ceErr)
```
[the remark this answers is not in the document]
Allyson-English: updated to not check nil!

internal/controllers/sacd_helpers.go
Allyson-English: that makes sense to me, I've merged this in

internal/controllers/sacd_helpers.go
Allyson-English: it's no longer load-bearing (the way it was written before some of the optimizations, it was load bearing but I like the way it's evolved now)

internal/controllers/token_exchange.go
Allyson-English: merged this in!

internal/controllers/token_exchange.go
Allyson-English: good with this one too, merged!

Allyson-English:
@elffjs I know you already approved but want to request one final review because I responded to your comments and merged in the PR's you'd had 
@KevinJoiner rerequesting your review as well 

I know there's a lot of stuff happening rn so no pressure on this for today! just wanted to get this request in to keep it moving, happy to merge in at a less chaotic time though

internal/controllers/sacd_helpers.go
elffjs: Sorry, I think I was unclear. Here's a diff:

https://github.com/DIMO-Network/token-exchange-api/compare/ex-3544-sacd-extended-attestation-permissions...annoying_point

The text of the error is slightly different in the test "Fail: permission not granted, address must match exactly", but we still get an error and everything else that used to pass still passes without change.

elffjs: A future project, I think, is to make sure the errors returned to the client are as specific as possible. We're bubbling up the result of `Error()` right now which is way better than nothing, but probably we can do better.

PR #51, opened 2025-05-23
Put developer license into JWT

elffjs:


Allyson-English (approved):

internal/services/dex_service.go
```diff
@@ -30,6 +30,7 @@ type PrivilegeTokenDTO struct {
 	NFTContractAddress string
 	Audience           []string
 	CloudEvents        *tokenclaims.CloudEvents
+	ResultSubject      string
```
KevinJoiner, line 33: nit: Could this just be called `Subject`? `ResultSubject` seems like a special term for the middleware only

KevinJoiner:
ref EX-3600

internal/services/dex_service.go
elffjs: I changed it to "response". That's closer to the `net/http` terminology.

internal/middleware/valid_dev_license.go
```diff
@@ -62,16 +71,31 @@ func NewDevLicenseValidator(idSvc IdentityService, logger zerolog.Logger) fiber.
...
 }
+
+// GetResultSubject returns the checksummed address of the developer license making
+// the request. The only exception occurs when the request emanates from the mobile map,
+// in which case the result is "dimo-driver".
+func GetResultSubject(c *fiber.Ctx) string {
+	addr, ok := c.Locals(resultSubjectKey).(string)
+	if !ok {
+		// This branch existing is a bit embarrassing.
+		return mobileAppAudience
```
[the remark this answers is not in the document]
elffjs: Okay, after much suffering I think we're good.

internal/controllers/token_exchange.go
```diff
@@ -157,18 +158,23 @@ func (t *TokenExchangeController) ExchangeToken(c *fiber.Ctx) error {
 }
 
 // Helper function to create and return the token
-func (t *TokenExchangeController) createAndReturnToken(c *fiber.Ctx, tokenReq *TokenRequest, ethAddr common.Address) error {
+func (t *TokenExchangeController) createAndReturnToken(c *fiber.Ctx, tokenReq *TokenRequest) error {
```
elffjs, line 161: We never used `ethAddr` anywhere, and once developer licenses are the only game in town then it becomes completely redundant.

elffjs:
I should have shown that it works. That's my bad, I'll add that here even though it's merged.

PR #54, opened 2025-05-27
refactor: update token ID handling to use big.Int

KevinJoiner:
- Change Claim TokenID type from string to *big.Int across the codebase
- Add ChainID support for future token validation
- Update NFT contract address handling to use common.Address type
- Add validation for required fields in SignPrivilegePayload


elffjs:
Keep in mind that https://github.com/DIMO-Network/token-exchange-api/pull/37, which may cause conflicts, isn't merged yet.

Can you give that one another review, actually?

KevinJoiner:
Going to rebase this after your dev_license PR is in as well just wanted to get something up.

PR #65, opened 2025-07-02
ex-3726-token-exchange-return-403-for-missing-perm

Allyson-English:
- when an address lacks an old-style permission, return 403 (forbidden) rather than 400
- change an error from `evaluateCloudEvents` and `evaluatePermissions` to also return 403 because if we are failing here it is because of a missing perm


elffjs:
Looking.

PR #82, opened 2025-08-26
Pr 3856 retrieve template doc cache ipfs

LorranSutter:


internal/ipfsdoc/ipfsdoc.go
```diff
@@ -0,0 +1,57 @@
...
+//
+// Parameters:
+//   - ctx: The context for the IPFS request, which can be used for cancellation and timeouts
+//   - source: The IPFS content identifier (CID) for the SACD document, typically with an "ipfs://" prefix
+//   - ipfsClient: The client used to fetch documents from IPFS
+//
+// Returns:
+//   - *cloudevent.RawEvent: A pointer to the parsed raw cloud event if valid
+//   - error: An error if the document could not be fetched, parsed, or doesn't have the correct type
+func GetValidSacdDoc(ctx context.Context, source string, ipfsClient IPFSClient) (*cloudevent.RawEvent, error) {
```
KevinJoiner, line 30: I think this code should live with the IPFS service as a member function. Then others can just use an interface for
GetValidSacdDoc instead of one for Fetch 
```go
 (i *IPFSClient) GetValidSacdDoc(...) {
      sacdDoc, err := i.Fetch(ctx, source)
      ...
  }
  ```

File: internal/signature/validator.go [hunk 2]
```diff
@@ -0,0 +1,105 @@
...
+	"fmt"
+
+	"github.com/DIMO-Network/token-exchange-api/internal/contracts/erc1271"
+	"github.com/ethereum/go-ethereum/accounts"
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/crypto"
+)
+
+var ERC1271MagicValue = [4]byte{0x16, 0x26, 0xba, 0x7e}
```
KevinJoiner, line 16: I don't think this needs to be exported. Which removes the ability for other packages to change the value.
```suggestion
var erc1271MagicValue = [4]byte{0x16, 0x26, 0xba, 0x7e}
```

internal/signature/validator.go
```diff
@@ -0,0 +1,105 @@
...
+
+func (f *defaultErc1271Factory) NewErc1271(address common.Address, backend bind.ContractBackend) (Erc1271Interface, error) {
+	return erc1271.NewErc1271(address, backend)
+}
+
+// Validator handles signature validation for both EOA and contract accounts
+type Validator struct {
+	backend bind.ContractBackend
+	// I don't like this, but it's the only way to get the mock to work.
+	erc1271Mgr erc1271Mgr
```
KevinJoiner, line 40: yeah I have had to do something similar. I wish there was a better way

internal/signature/validator.go
```diff
@@ -0,0 +1,105 @@
...
+	IsValidSignature(opts *bind.CallOpts, hash [32]byte, signature []byte) ([4]byte, error)
+}
+
+type erc1271Mgr interface {
+	NewErc1271(address common.Address, backend bind.ContractBackend) (Erc1271Interface, error)
+}
+
+type Erc1271Factory interface {
+	NewERC1271(address common.Address, backend bind.ContractBackend) (Erc1271Interface, error)
+}
```
KevinJoiner, line 28: I think we can remove this one

internal/services/template/template.go
```diff
@@ -0,0 +1,156 @@
...
+	templatePermGrants := s.extractPermissionsFromAgreements(data.Agreements, assetDID)
+	return templatePermGrants, nil
+}
+
+func (s *Service) extractPermissionsFromAgreements(agreements []models.TemplateAgreement, assetDID cloudevent.ERC721DID) map[string]bool {
+	templatePermGrants := make(map[string]bool)
+
+	for _, agreement := range agreements {
+		// TODO(lorran) agreement.Asset does not have :id, fix this if needed
+		if agreement.Asset != assetDID.String() {
```
KevinJoiner, line 127: I think we want to check that it matches the address.
I don't think we need to error
```suggestion
		if agreement.Asset != assetDID.ContractAddress.Hex(){
```

KevinJoiner (changes requested):

internal/app/app.go
```diff
@@ -9,11 +9,13 @@ import (
 	"github.com/DIMO-Network/shared/pkg/middleware/metrics"
 	"github.com/DIMO-Network/token-exchange-api/internal/config"
 	"github.com/DIMO-Network/token-exchange-api/internal/contracts/sacd"
+	"github.com/DIMO-Network/token-exchange-api/internal/contracts/template"
 	"github.com/DIMO-Network/token-exchange-api/internal/controllers/httpcontroller"
 	"github.com/DIMO-Network/token-exchange-api/internal/controllers/rpc"
 	"github.com/DIMO-Network/token-exchange-api/internal/middleware"
 	"github.com/DIMO-Network/token-exchange-api/internal/services"
 	"github.com/DIMO-Network/token-exchange-api/internal/services/access"
+	template_service "github.com/DIMO-Network/token-exchange-api/internal/services/template"
```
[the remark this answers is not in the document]
LorranSutter: I have to say the `_` was bothering me, but I saw it right below and thought it wouldn't be a problem

internal/signature/validator.go
LorranSutter: I think I was using this var in another place before. Makese sense

internal/services/template/template.go
LorranSutter: Ah, this was changed in this PR https://github.com/DIMO-Network/token-exchange-api/pull/83

internal/ipfsdoc/ipfsdoc.go
LorranSutter: Makes perfect sense

LorranSutter:
Addressed all the comments in https://github.com/DIMO-Network/token-exchange-api/pull/83

PR #83, opened 2025-08-27
Pr 3857 template activation check on chain

LorranSutter:


internal/autheval/autheval.go
```diff
@@ -147,13 +148,53 @@ func EvaluateIDsByGrantSource(globalGrants *set.StringSet, sourceGrants *set.Str
...
+		}
+	} else {
+		// Template is not active: only additional SACD permissions are granted
+		// Permissions in template that match SACD are NOT granted
+		for sacdPerm, granted := range sacdPermissions {
+			if granted && !templateResult.Permissions[sacdPerm] {
+				resultPermissions[sacdPerm] = true
+			}
+		}
+	}
```
KevinJoiner, line 182: I think we should match the behavior for the permission bits on chain. I believe their if the template is not active then we hasPermissions returns false without checking any other permissions.

internal/models/permission_record.go
```diff
@@ -34,7 +34,7 @@ type Agreement struct {
 	IDs                  []string     `json:"ids"`
 	EffectiveAt          time.Time    `json:"effectiveAt"`
 	ExpiresAt            time.Time    `json:"expiresAt"`
-	PermissionTemplateID string       `json:"permissionTemplateID"`
+	PermissionTemplateID string       `json:"permissionTemplateId"`
```
KevinJoiner, line 37: Sorry I didn't catch this before but shouldn't the Permission Template be higher up on the SACDData struct. Since on chain the `templateId` is set on the permission record not each individual permission bit. I am worried that it will hurts us in the future if we have different behavior based on if you specify a template on chain vs specify a template in the SACD doc? 
This might need further discussion but, I want to make sure we are all on the same page.

internal/services/template/template.go
```diff
@@ -115,17 +120,51 @@ func (s *Service) GetTemplatePermissions(ctx context.Context, permissionTemplate
...
 	templatePermGrants := make(map[string]bool)
 
+	// Create the base DID without the token ID for comparison
+	baseDID := fmt.Sprintf("did:erc721:%d:%s", assetDID.ChainID, assetDID.ContractAddress.Hex())
+
 	for _, agreement := range agreements {
-		// TODO(lorran) agreement.Asset does not have :id, fix this if needed
-		if agreement.Asset != assetDID.String() {
-			// TODO(lorran) Consider if we want an error here
+		if agreement.Asset != baseDID {
```
KevinJoiner, line 167: I don't think the asset on chain should look like `did:erc721:<chain_id>:<contarct_address>`
either we just use the 0x address for the asset on chain or we should use ethr did https://github.com/DIMO-Network/cloudevent?tab=readme-ov-file#ethereum-did

test.go
```diff
@@ -0,0 +1,18 @@
+package main
```
KevinJoiner, line 1: Probably committed on accident

internal/models/permission_record.go
LorranSutter: Yeah... we can't have this level of granularity on chain. Let's change that in another PR if we go this way

KevinJoiner (changes requested):

PR #118, opened 2026-06-10
feat(access): account-owner self-grant for account-level assets

zer0stars:
## What

`ValidateAccess` now grants requested permissions and cloud-event filters outright when the requester's wallet **is** the account-level asset (`did:ethr:<chainId>:<address>`). No grantor/grantee relationship exists to validate in that case.

## Why

Personal (account-subject) documents are landing in DIS with `subject: did:ethr:...`. To read them back through fetch-api, the mobile app exchanges the user's JWT for a token scoped to the user's own account DID. Without this change that exchange requires an on-chain self-referential SACD record (`AccountPermissionRecords(wallet, wallet)`), which no user has and which would be pointless friction — you always control your own account's data.

Mirrors the existing vehicle owner auto-grant in the legacy bits path (`sacdproxy.GetPermissions`).

## Notes

- Non-self callers on account assets are unchanged: SACD source-doc path, then legacy `GetAccountPermissions` fallback.
- Two account-asset test fixtures used the asset wallet itself as the caller while asserting the record-lookup path; they now use a distinct caller so grantee validation stays covered.
- New test: self-grant for permissions + event filters, plus a non-self caller still hitting the record path.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ VERSION   := $(shell git describe --tags || echo "v0.0.0")
 VER_CUT   := $(shell echo $(VERSION) | cut -c2-)
 
 # Dependency versions
-GOLANGCI_VERSION   = latest
+GOLANGCI_VERSION   = v2.12.1
 PROTOC_VERSION		= 31.1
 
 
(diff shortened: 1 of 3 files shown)


zer0stars:
Closing — superseded by the SACD-only access decision: personal-document reads always go through the user's on-chain account SACD grant to the app's dev license (token-exchange's existing non-self path), so nothing consumes the self-grant. The CI fix from this branch lives on in #119.

